PR #26, opened 2018-04-05
Add Open Graph image

petk:
Hello, this patch adds the [Open Graph image](http://ogp.me) to the main layout.

How this looks like on Facebook:
doctrine-og

Size is picked with 1200x1200 px and the logo at 610px because the FB will crop this image to 1200x630 px on a usual share and scale it to a square when commenting link as pointed on above image.

Thanks for considering checking this out as well.


source/_layouts/layout.html
```diff
@@ -46,6 +46,10 @@
         <meta name="msapplication-wide310x150logo" content="{{ site.url }}/images/mstile-310x150.png" />
         <meta name="msapplication-square310x310logo" content="{{ site.url }}/images/mstile-310x310.png" />
 
+        <meta property="og:image" content="{{ site.url }}/images/og.png" />
```
Ocramius: Can the `og.png` also just be `og.svg`?

PR #49, opened 2018-04-07
Allow line break in menu item

lex111:
Before
image

After
image


jwage:
I am actually not sure how I feel about this. I think it looks worse without the line height and the wrapping. I am working on fixing the really long section titles too that would eliminate the need for the wrapping.

jwage:
I am going to work on a change to introduce the wrapping but keep the padding that the line height was giving.

jwage:
Pulled in your changes here #52 and will work on it there.

PR #74, opened 2018-04-09
Fix issue with rst syntax for bulleted lists. Fixes #71

### LINKED ISSUE #71: Bullet lists are not rendered correctly
https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/1.2/reference/annotations-reference.html#title.1.47

vs

http://doctrine-orm.readthedocs.io/projects/doctrine-mongodb-odm/en/1.2.x/reference/annotations-reference.html#referencemany

jwage:


jwage:
This is only a hack. I am rethinking using this gregwar/rst library. It is not very complete. I might look at using an rst parser from python or something. I don't think PHP has any RST parsers that will be good enough.

PR #75, opened 2018-04-10
Remove 1.1 from ODM docs

malarzm:
1.1 is not supported and anyway, docs for 1.2 are sufficient

### DIFF
--- a/app/config/projects.yml
+++ b/app/config/projects.yml
@@ -92,10 +92,6 @@ parameters:
                     branchName: 1.2.x
                     slug: 1.2
                     current: true
-                -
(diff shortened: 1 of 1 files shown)


alcaeus:
Looks good - although we may just want to add a note that the version is no longer supported. Might be useful to keep the docs around for reference (Symfony does it this way).

jwage:
Instead of removing, add maintained: false and add  older versions that aren't maintained too. I'll update the UX to incorporate if the version is maintained or not.

malarzm:
Closing in favour of #77

PR #80, opened 2018-04-10
Twitters & websites

Majkl578:


app/config/team_members.yml
```diff
@@ -26,9 +26,10 @@ parameters:
             active: true
         carusogabriel:
             name: 'Gabriel Caruso'
+            twitter: carusogabriel
             github: carusogabriel
             avatarUrl: 'https://avatars3.githubusercontent.com/u/16328050?v=4'
-            website: 'https://twitter.com/carusogabriel'
+            website: ''
```
Ocramius: `~`?

app/config/team_members.yml
```diff
@@ -66,14 +69,15 @@ parameters:
         jwage:
             name: 'Jonathan H. Wage'
             twitter: jwage
-            website: 'http://www.twitter.com/jwage'
+            website: ''
```
Ocramius: `~`?

app/config/team_members.yml
```diff
@@ -217,9 +222,10 @@ parameters:
             location: null
         FabioBatSilva:
             name: 'Fabio B. Silva'
+            twitter: FabioBatSilva
             github: FabioBatSilva
             avatarUrl: 'https://avatars2.githubusercontent.com/u/588172?v=4'
-            website: 'https://twitter.com/FabioBatSilva'
+            website: ''
```
Ocramius: `~`

app/config/team_members.yml
```diff
@@ -110,7 +115,7 @@ parameters:
             twitter: malarzm
             github: malarzm
             avatarUrl: 'https://avatars1.githubusercontent.com/u/4947711?v=4'
-            website: 'http://twitter.com/malarzm'
+            website: ''
```
Ocramius: `~`

app/config/team_members.yml
Majkl578: Empty string is used everywhere else so I used the same. @jwage any reason for `''`?

jwage: That was just from the github import. Null is fine.

PR #88, opened 2018-04-10
Enhance the sidebar functionality so projects can provide an explicit…

jwage:
… sidebar.rst to use instead of autogenerating.


app/src/Doctrine/Website/DoctrineSculpinBundle/Directive/ToctreeDirective.php
```diff
@@ -18,34 +19,26 @@ public function process(Parser $parser, $node, $variable, $data, array $options)
         $kernel = $parser->getKernel();
         $files = array();
 
-        // This sucks. I don't even think it works 100% right.
```
jwage: It still sucks but it is better and I am more confident now in how it should work and how it works now. I am going to keep on working on this as I move through the documentation for all the projects.

tests/Doctrine/Website/Tests/Docs/resources/example-project/docs/en/index.rst
```diff
@@ -66,3 +66,37 @@ Reference Anchor
...
+TOC
+---
+
+.. tocheader:: TOC Title
+
+.. toc::
+
+.. toctree::
+    :depth: 3
+    :glob:
```
deeky666: @jwage this does not make any sense to me. Can you please explain why you use glob here when you build the TOC explicitly afterwards? Is it necessary?

jwage: This is a test for the ToctreeDirective which supports glob wildcards for building table of contents. We still use this in other places in the docs and if a project does not provide a sidebar.rst it will fall back to generating one automatically.

app/src/Doctrine/Website/DoctrineSculpinBundle/Directive/ToctreeDirective.php
jwage: I should also add that I'm going to move ToctreeDirective to doctrine/rst fork and move the associated tests there.

tests/Doctrine/Website/Tests/Docs/resources/example-project/docs/en/index.rst
deeky666: @jwage sure, but you have glob and explicit TOC declaration together in this example. The example before already uses glob. This is what confuses me here

jwage: Oh sorry, I misunderstood. You are right, the glob option can be removed here since I do not pass any paths with a `*` in them.

PR #109, opened 2018-04-14
Fix highlighting on the contribute page

stof:
The auto-detection of the language was selecting the wrong language for several code blocks on the page.

### DIFF
--- a/source/contribute.md
+++ b/source/contribute.md
@@ -56,13 +56,13 @@ Initial Setup
     project you want to contribute to.
 - Clone your fork locally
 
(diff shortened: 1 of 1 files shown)


jwage:
Thanks @stof 

I just now realized that my whole staging.doctrine-project.org setup won't work for pull requests from forks. It uses the push web hook on the repo. I also just realized that travis isn't running for pull requests from forks either.

jwage:
I had it set to only run pushed branches. I changed it to pull requests.

PR #141, opened 2018-04-24
Add ability to easily use google translate from the UI.

### LINKED ISSUE #140: Add easy google translate button
We can do this with javascript and make it pretty seamless. We need to make it so the code blocks do not get translated.

jwage:
Fixes #140

### DIFF
--- a/source/_layouts/layout.html
+++ b/source/_layouts/layout.html
@@ -110,7 +110,7 @@
     </head>
     <body>
         {% block navbar %}
-            <nav class="navbar navbar-expand-md navbar-dark sticky-top bg-dark flex-md-nowrap">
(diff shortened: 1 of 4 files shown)


source/_layouts/layout.html
```diff
@@ -211,8 +221,16 @@
...
             new Search(projectSlug, versionSlug);
+
+            function googleTranslateElementInit() {
+                new google.translate.TranslateElement(
+                    {pageLanguage: 'en'}, 'google_translate_element'
+                );
+            }
         </script>
 
+        <script type="text/javascript" src="//translate.google.com/translate_a/element.js?cb=googleTranslateElementInit"></script>
```
stof: I would always use HTTPS instead of making it protocol-relative. there is no reason to use HTTP when HTTPS is available.

PR #176, opened 2018-06-14
Add the Doctrine Coding Standard project to the website.

jwage:
Needs to be merged after I add docs to the master and 4.0 branch in doctrine/coding-standard. /cc @carusogabriel


app/config/projects.yml
```diff
@@ -462,6 +462,32 @@ parameters:
...
+            docsRepositoryName: coding-standard
+            docsPath: /docs
+            description: >
+                The Doctrine Coding Standard is a set of PHPCS rules applied to all Doctrine projects.
+
+            keywords: [coding, standard, rules, sniffs, checks]
+            versions:
+                -
+                    name: '5.0'
+                    branchName: 'master'
```
Ocramius: Wondering about the maintainability of this: would it make sense to use the `branch-alias` feature of composer to figure these out? To be more clear, composer only looks at `master` when finding branch aliases

app/config/projects.yml
```diff
@@ -462,6 +462,32 @@ parameters:
...
+                    slug: 'latest'
+                    upcoming: true
+                -
+                    name: '4.0'
+                    branchName: '4.0'
```
Ocramius: Hmm, the branch is called 4.x

app/config/projects.yml
jwage: It would be nice to have something more advanced here. But yes, currently as soon as we release 5.0 and master becomes 6.0, we'd have to change this from master to 5.x or whatever we name the branch.

PR #182, opened 2018-07-02
Fix of links on 404 page

SenseException:
On the 404 page are several links in an unordered list, where some of them were dead links itself. I adapted that list and fixed the links.


jwage:
Thanks

PR #189, opened 2018-07-13
Fix number of fixes/improvements and period since last release

jdreesen:
According to the release notes and tags of DBAL 2.8.0 it's more than 30 fixes/improvements instead of 20 in 3 months instead of 2.


Majkl578:
Cool, thanks! I think @morozov updated the numbers before publishing the release, but the blog post was already published by that time. 😅

morozov:
Doctrine is a geographically distributed organization with eventual consistency. We need to accept that.

Issue #190, opened 2018-07-30
Search weirdly broken (at least in Safari)


ondrejmirtes:
Hi,
I'm trying to search on the Doctrine website (https://www.doctrine-project.org) for string `matching`. It behaves in a weird way:

1) I can't scroll the results. In scrolls the background instead.
2) When I defocus the input, the search results disappear.

I have Safari 11.1.2 on macOS 10.13.6.

ondrejmirtes:
BTW it does the same thing in Chrome.

SenseException:
Having the result of the search as sticky as the navigation makes the search hard to use, even on mobiles. We could move the result container out of the `nav` tag as a quick win, but I'm not sure if this would still be a good markup.

jwage:
Thanks to @SenseException for making a change that improves this. It is still not "great", but it is better.

PR #199, opened 2018-08-08
Move search result container out of sticky area

SenseException:
This is a fix for #190.

This PR also handles the search results for mobiles, because the sticky option for the results are not usable for screens, that can't hold all results in the visible area and won't scroll further when reached the sites end.

### DIFF
--- a/source/_layouts/layout.html
+++ b/source/_layouts/layout.html
@@ -162,15 +162,15 @@
                         <!-- SearchBox widget will appear here -->
                     </div>
                 </div>
-
-                <div class="search-results rounded">
-                    <div id="hits">
-                        <!-- Hits widget will appear here -->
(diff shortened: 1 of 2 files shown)


Majkl578:
Looks a bit broken in Chrome dev tools: 


(I am unable to focus to / write into the search box in Firefox mobile preview).

SenseException:
Ah, the absolute position is still there. Thank you for the feedback.

jwage:
This design of the search really caused me a lot of grief. I couldn't find a good solution. Happy to see if you can make it better.

SenseException:
I've removed the position behaviour from the search result now.

In a future PR, we need to redesign the navigation. It takes too much space from the search results.

Issue #111, opened 2018-04-14
Add a place that documents and standardizes our policies for releases, BC, depreciation, etc.


jwage:
- Release cycle
  - Every month we have a bug fix release.
  - Every 6 months we have a minor release.
  - Every 2 years we have a major release.
- Only maintain at most 3 branches at a time per project.
  - 1 branch for current stable release 
  - 1 branch for previous stable release
  - 1 branch for master (next)
- All bug fixes should be targeted to the master branch and then cherry-picked to each branch where the bug fix applies.
  - We need some kind of tooling to automate this and ensure that all applicable bugs are cherry-picked to each branch where relevant.
- New features are merged in to the master branch.
- Trigger deprecations warnings at runtime whenever possible before the BC break is released.

ostrolucky:
Target branch for bug fixes contradicts https://github.com/doctrine/doctrine2/blob/master/CONTRIBUTING.md Also rebase strategy clearly in use is not mentioned.

@Majkl578 @guilhermeblanco can you correct this please?

Majkl578:
@ostrolucky That's intentional (see https://github.com/doctrine/doctrine2/pull/7107), ORM has non-standard workflow right now by design.

jwage:
I setup a new section on the website where we can start to publish these policies https://www.doctrine-project.org/policies.html

PR #201, opened 2018-08-12
Add small note about the cherry-pick workflow

greg0ire:
People should be encouraged to let their change flow from master to
lower branches, rather than targetting directly the release branch that
matches their setup.

### DIFF
--- a/source/contribute.md
+++ b/source/contribute.md
@@ -133,7 +133,9 @@ Topic branches are for contributors to develop bug fixes, new features,
 etc. so that they can be easily merged to **master**. They must follow a
 few simple rules as listed below:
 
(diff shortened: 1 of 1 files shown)


ostrolucky:
rel: https://github.com/doctrine/doctrine-website/issues/111

SenseException:
Should we mention differences (f.e. like in ORM) in there too?

greg0ire:
@SenseException looks like the ORM has its own contributing guide: https://github.com/doctrine/doctrine2/blob/master/CONTRIBUTING.md

SenseException:
Yes and we should mention that in there as long as there are custom guides. We don't need to mention details. just mention them so that contributors will follow them when it comes to a PR.

PR #206, opened 2018-08-22
Explicitly set search box color to black

Majkl578:
Without this, the input has white text on white background with dark GTK+ theme (Gnome).

### DIFF
--- a/source/css/style.css
+++ b/source/css/style.css
@@ -61,6 +61,8 @@ h1, h2, h3, h4, h5, h6 {
 .navbar input {
     height: 40px;
     margin-top: 2px;
+    color: black;
+    background-color: white;
 }
 
 .navbar a:hover {


Ocramius:
@Majkl578 I have the same issue with the dark GTK+ theme too - don't think this is the website's fault...

Majkl578:
It may not be necessarily our fault directly, but one of our CSS dependencies does that - `.ais-search-box--input` class declares `background: #fff`. Normally dark GTK+ theme sets dark input background with white text.
So default dark GTK+ theme makes it look like:

Majkl578:
it's Algolia instantsearch.js: https://github.com/algolia/instantsearch.js/blob/v2.10.1/src/css/theme/_search-box.scss#L12

Issue for that: https://github.com/algolia/instantsearch.js/issues/3075

jwage:
:+1: Can you rebase and I will merge?

SenseException:
With changes in Instantsearch in the background color, we will run into similar problems by just adding color black. If this is fixed in this PR, we also need a background color to prevent future off-coloring.

jwage:
So we need to also add our own white background here too?

SenseException:
I suggest that we do that. I also suggested to use background and foreground color on algolia/instantsearch.js#3075. If we don't use our own colors for a third part library, we might run into similar problems when changes happen on their side.

SenseException:
@Majkl578 Could you please add the background-color so we don't rely to the one of instantsearch.js?

jwage:
I can finish this PR. I will push up the other change.

SenseException:
My algolia/instantsearch.js PR got merged too. If you see that a background-color or color is missing, add it into the project code.

PR #207, opened 2018-08-23
Remove sculpin and replace with something simpler.

jwage:


lib/Builder/SourceFileRepository.php
```diff
@@ -40,9 +40,9 @@ public function getFiles(string $buildDir, string $inPath = 'source') : array
         $files = [];
 
         foreach ($finder as $splFileInfo) {
-            $path = $splFileInfo->getRealPath();
+            $path = (string) $splFileInfo->getRealPath();
```
Ocramius: These should probably be assertions rather than casts

lib/Builder/SourceFileRepository.php
```diff
@@ -40,9 +40,9 @@ public function getFiles(string $buildDir, string $inPath = 'source') : array
...
 
-            $contents = file_get_contents($path);
+            $contents = (string) file_get_contents($path);
```
Ocramius: These should probably be assertions rather than casts

lib/Deployer.php
```diff
@@ -122,6 +122,6 @@ private function getFileContents(string $file) : string
             return '';
         }
 
-        return trim(file_get_contents($file));
+        return trim((string) file_get_contents($file));
```
Ocramius: These should probably be assertions rather than casts

lib/Docs/CodeBlockWithLineNumbersRenderer.php
```diff
@@ -107,7 +107,7 @@ private function generateLineNumbersTableTemplate(array $lines, string $codeElem
         $lineTableRows = [];
 
         foreach ($lines as $key => $line) {
-            $lineNumber = $key + 1;
+            $lineNumber = (int) $key + 1;
```
Ocramius: This should be a documentation on `$lines`: if you document it as `@var array <int, Foo>` it will be OK. A cast is not necessary

lib/Docs/RSTBuilder.php
```diff
@@ -141,7 +141,7 @@ private function copyRst(Project $project, ProjectVersion $version) : void
         // check if we have an explicit sidebar file to use
         // otherwise just use the default autogenerated sidebar
         $sidebarPath = $docsPath . '/sidebar.rst';
-        $sidebar     = file_exists($sidebarPath) ? file_get_contents($sidebarPath) : self::DEFAULT_SIDEBAR;
+        $sidebar     = file_exists($sidebarPath) ? (string) file_get_contents($sidebarPath) : self::DEFAULT_SIDEBAR;
```
Ocramius: Same here, assertions. Maybe make your own `file_get_contents()` that throws and always returns `string`?

lib/Docs/RSTBuilder.php
```diff
@@ -152,7 +152,7 @@ private function copyRst(Project $project, ProjectVersion $version) : void
 
             $this->ensureDirectoryExists(dirname($newPath));
 
-            $content = trim(file_get_contents($file));
+            $content = trim((string) file_get_contents($file));
```
Ocramius: Same as above: assertion rather than cast

lib/Docs/RSTBuilder.php
```diff
@@ -276,7 +276,7 @@ private function removeMetaFiles(string $path) : void
 
     private function getProjectDocsPath(Project $project) : string
     {
-        return realpath($project->getAbsoluteDocsPath($this->projectsPath));
+        return (string) realpath($project->getAbsoluteDocsPath($this->projectsPath));
```
Ocramius: Same as above: assertion rather than cast

lib/Projects/ProjectRepository.php
```diff
@@ -8,7 +8,7 @@
 
 class ProjectRepository
 {
-    /** @var Project[] */
+    /** @var mixed[][] */
```
Ocramius: Is there something better than `mixed` that we can use here?

lib/RST/Directive/CodeBlockDirective.php
```diff
@@ -53,13 +55,13 @@ public function process(
...
     ) : void {
-        if (! $node) {
+        if (! $node instanceof CodeNode) {
             return;
         }
 
         $kernel = $parser->getKernel();
 
-        $lines = $this->getLines($node->getValue());
+        $lines = $this->getLines((string) $node->getValue());
```
Ocramius: Assertion rather than cast

lib/RST/Directive/CodeBlockDirective.php
```diff
@@ -84,7 +86,9 @@ public function process(
      */
     private function getLines(string $code) : array
     {
-        $lines = array_reverse(preg_split('/\r\n|\r|\n/', $code));
+        $lines = preg_split('/\r\n|\r|\n/', $code);
+        $lines = is_array($lines) ? $lines : [];
```
Ocramius: This is another cast that is to be an assertion instead

lib/RST/Directive/CodeBlockDirective.php
```diff
@@ -84,7 +86,9 @@ public function process(
...
+        $lines = array_reverse($lines);
```
Ocramius: Use a new variable instead of overwriting the same one

lib/Twig/MainExtension.php
```diff
@@ -77,7 +77,7 @@ public function getApiDocsUrls() : array
      */
     private function getUrlsFromFiles(string $path, string $extension = 'html') : array
     {
-        $root = realpath(__DIR__ . '/../../source');
+        $root = (string) realpath(__DIR__ . '/../../source');
```
Ocramius: assertions

lib/Twig/MainExtension.php
```diff
@@ -107,8 +107,8 @@ private function getUrlsFromFiles(string $path, string $extension = 'html') : ar
 
     private function getAssetCacheBuster(string $path) : string
     {
-        $assetPath = realpath(__DIR__ . '/../../source/' . $path);
+        $assetPath = (string) realpath(__DIR__ . '/../../source/' . $path);
```
Ocramius: assertions

lib/Twig/MainExtension.php
```diff
@@ -107,8 +107,8 @@ private function getUrlsFromFiles(string $path, string $extension = 'html') : ar
...
 
-        return substr(md5(file_get_contents($assetPath)), 0, 6);
+        return substr(md5((string) file_get_contents($assetPath)), 0, 6);
```
Ocramius: assertions

Ocramius:
As discussed with @jwage, shipping this, and the diff size is expected

SenseException:
Joy. Now my spent time for code review was for naught.

Majkl578:
@SenseException I feel with you. `+28,341 −54,204` and I haven't even seen this PR until now. It was open for <24h.

PR #208, opened 2018-08-25
Show which project and version you are searching in the search box placeholder

jwage:
screen shot 2018-08-24 at 11 17 45 pm


jwage:
@SenseException I thought this would be a nice little addition after we chatted in Gitter about this functionality being unknown to the end user.

PR #211, opened 2018-08-26
Rely less on twig extensions by setting up initial controllers.

jwage:
I am not married to any of the naming, organization or concepts introduced in this PR. I am happy to tweak naming or anything based on feedback. The goal was to make the data that Twig templates rely on much more explicit and to not rely on Twig extensions to provide data for the different templates.


SenseException:
I get an Exception on `./bin/console build-website`:  The merge filter only works with arrays or "Traversable", got "NULL" as first argument in "50d540afc525a4b55b2a2f5704ca11434846984b" at line 81

This Twig_Error_Runtime is thrown in vendor/twig/twig/lib/Twig/Extension/Core.php at line 550.

lib/Controller/ControllerExecutor.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function execute(SourceFile $sourceFile) : array
+    {
+        [$className, $methodName] = $sourceFile->getParameter('controller');
+
+        $controller = $this->controllerFactory->create($className);
+
+        $reflectionMethod = (new ReflectionClass($controller))->getMethod($methodName);
+
+        $controllerResult = $reflectionMethod->invokeArgs($controller, [$sourceFile]);
```
SenseException: Only 3 of the currently 9 controller actions need `$sourceFile` injected. In the other controller actions it is an unused argument. I'm not sure about that approach. Why did you choose the way of ControllerFactory, ControllerExecutor, ControllerResult and controller methods having the need to get the same `SourceFile` argument?

lib/Controller/ControllerFactory.php
```diff
@@ -0,0 +1,39 @@
...
+use function get_class;
+use function sprintf;
+
+class ControllerFactory
+{
+    /** @var object[] */
+    private $controllers;
+
+    /**
+     * @param mixed[] $controllers
```
SenseException: Isn't this `object[]` too like in the property?

lib/Controller/ControllerFactory.php
```diff
@@ -0,0 +1,39 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Controller;
+
+use InvalidArgumentException;
+use function get_class;
+use function sprintf;
+
+class ControllerFactory
```
SenseException: This class is less a factory but more a provider of already created instances. Even when only one of them is needed.

lib/Controller/ControllerFactory.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /** @var object[] */
+    private $controllers;
+
+    /**
+     * @param mixed[] $controllers
+     *
+     * @return mixed[]
```
SenseException: No return needed.

lib/Sitemap/SitemapPageRepository.php
```diff
@@ -0,0 +1,70 @@
...
+
+        if (! file_exists($path)) {
+            return [];
+        }
+
+        $it = new RecursiveDirectoryIterator($path, RecursiveDirectoryIterator::SKIP_DOTS);
+
+        $urls = [];
+        foreach (new RecursiveIteratorIterator($it) as $file) {
+            $path = (string) $file;
```
SenseException: Using `RecursiveDirectoryIterator::CURRENT_AS_PATHNAME` won't create a SplFileInfo but a path name.

jwage:
@SenseException I pushed up two new commits to make things more strict so maybe we can get a better error message when you run `./bin/console build-website`. Try running a fresh `./bin/console build-docs --sync-git` if you still get an error (let me know what it is) and see if that clears it up.

lib/Blog/BlogPost.php
```diff
@@ -0,0 +1,74 @@
...
+    /** @var DateTime */
+    private $date;
+
+    public function __construct(
+        string $url,
+        string $title,
+        string $authorName,
+        string $authorEmail,
+        string $contents,
+        DateTime $date
```
Majkl578: Please always use `DateTimeImmutable`.

lib/Blog/BlogPostRepository.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * @return BlogPost[]
+     */
+    public function findAll() : array
+    {
+        $sourceFiles = $this->sourceFileRepository->getFiles('', 'source/blog');
+
+        usort($sourceFiles, function (SourceFile $a, SourceFile $b) {
```
Majkl578: closure needs a return type

lib/Blog/BlogPostRepository.php
```diff
@@ -0,0 +1,52 @@
...
+            return $a->getDate()->getTimestamp() - $b->getDate()->getTimestamp();
+        });
+
+        $reversedSourceFiles = array_reverse($sourceFiles);
+
+        return array_map(function (SourceFile $sourceFile) {
```
Majkl578: closure needs a return type

lib/Builder/SourceFile.php
```diff
@@ -56,7 +53,7 @@ public function getWritePath() : string
 
     public function getUrl() : string
     {
-        return $this->parameters['url'];
+        return (string) $this->parameters->getParameter('url');
     }
 
     public function getDate() : DateTime
```
Majkl578: `DateTimeImmutable`

lib/Builder/SourceFileRenderer.php
```diff
@@ -29,26 +37,36 @@ public function render(SourceFile $sourceFile, string $contents) : string
...
+        return $this->twigRenderer->render($template, $parameters);
     }
 
     /**
      * @return mixed[]
      */
     private function preparePageParameters(SourceFile $sourceFile) : array
     {
-        return array_merge($sourceFile->getParameters(), [
+        return array_merge($sourceFile->getParameters()->getAll(), [
```
Majkl578: Can we use `+` instead of `array_merge()` here? Should be ok for assoc array.

lib/Builder/SourceFileRenderer.php
```diff
@@ -29,26 +37,36 @@ public function render(SourceFile $sourceFile, string $contents) : string
...
+        $parameters = [
             'page' => $pageParameters,
             'site' => $this->site,
-        ]);
+        ];
+
+        if (isset($parameters['page']['controller'])) {
+            $controllerParameters = $this->controllerExecutor->execute($sourceFile);
+
+            $parameters = array_merge($parameters, $controllerParameters);
```
Majkl578: Can we use + instead of array_merge() here? Should be ok for assoc array.

lib/Builder/SourceFileRenderer.php
```diff
@@ -29,26 +37,36 @@ public function render(SourceFile $sourceFile, string $contents) : string
...
+        return array_merge($sourceFile->getParameters()->getAll(), [
             'date' => $sourceFile->getDate(),
         ]);
     }
 
     private function prepareTemplate(SourceFile $sourceFile, string $contents) : string
     {
         if ($sourceFile->isLayoutNeeded()) {
-            if (preg_match_all('/{%\s+block\s+(\w+)\s+%}(.*?){%\s+endblock\s+%}/si', $contents, $matches) === 0) {
+            $hasContentBlock = preg_match_all('/{%\s+block\s+(\w+)\s+%}(.*?){%\s+endblock\s+%}/si', $contents, $matches) === 0;
```
Majkl578: Is this variable name correct? From `=== 0` I assume it should rather be `$doesNotHaveContentBlock`.
Also when `$contents === ''`, this regex doesn't need to run (condition below will be false anyway).

lib/Commands/ClearBuildCacheCommand.php
```diff
@@ -0,0 +1,94 @@
...
+            ->setDescription('Clear the build cache.')
+            ->addArgument(
+                'build-dir',
+                InputArgument::OPTIONAL,
+                'The directory where the build repository is cloned.'
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output) : void
```
Majkl578: Correct return type for `execute()` is (unsigned) `int`.

lib/Commands/ClearBuildCacheCommand.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        $buildDir = $input->getArgument('build-dir');
+
+        if ($buildDir === null) {
+            $buildDir = sprintf('%s/build-%s', $this->rootDir, $this->env);
```
Majkl578: You can put this directly as a default value to `->addArgument()` inside `configure()` (4th parameter IIRC).

lib/Controller/ControllerExecutor.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->controllerProvider = $controllerProvider;
+    }
+
+    /**
+     * @return mixed[]
+     */
+    public function execute(SourceFile $sourceFile) : array
+    {
+        [$className, $methodName] = $sourceFile->getParameter('controller');
```
Majkl578: (Marked these Scrutinizer errors as false positives.)

lib/Controller/ControllerProvider.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public function __construct(array $controllers)
+    {
+        foreach ($controllers as $controller) {
+            $this->controllers[get_class($controller)] = $controller;
+        }
+    }
+
+    /**
+     * @return object
```
Majkl578: What PHP version do we require for building website? `composer.json` doesn't specify one, this should be a native return type as of 7.2.

lib/Projects/Project.php
```diff
@@ -192,11 +194,20 @@ public function getUnmaintainedVersions() : array
...
     }
 
-    public function getVersion(string $slug) : ?ProjectVersion
+    /**
+     * @throws InvalidArgumentException
+     */
+    public function getVersion(string $slug) : ProjectVersion
     {
-        return $this->getVersions(function (ProjectVersion $version) use ($slug) {
+        $projectVersion = $this->getVersions(function (ProjectVersion $version) use ($slug) {
```
Majkl578: Closure needs a return type.

lib/Projects/ProjectRepository.php
```diff
@@ -30,15 +33,21 @@ public function findOneBySlug(string $slug) : Project
...
+        throw new InvalidArgumentException(sprintf('Could not find Project with slug "%s"', $slug));
     }
 
     /**
      * @return Project[]
      */
     public function findAll() : array
     {
-        return array_map(function (array $project) {
+        $projects = array_map(function (array $project) {
```
Majkl578: Closure needs a return type.

lib/Sitemap/SitemapPage.php
```diff
@@ -0,0 +1,32 @@
...
+
+class SitemapPage
+{
+    /** @var string */
+    private $url;
+
+    /** @var DateTime */
+    private $date;
+
+    public function __construct(string $url, DateTime $date)
```
Majkl578: DateTimeImmutable

lib/Controller/ControllerProvider.php
SenseException: I would prefer 7.2 too.

jwage: 7.1 currently.

PR #212, opened 2018-08-30
Add more unit tests.

jwage:


tests/Builder/SourceFileParametersTest.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Tests\Builder;
+
+use Doctrine\Website\Builder\SourceFileParameters;
+use PHPUnit\Framework\TestCase;
+
+class SourceFileParametersTest extends TestCase
```
SenseException: Please add a test where an unknown parameter is called with `getParameter()` and we expect `null`. This is uncovered code.

tests/Builder/SourceFileRendererTest.php
```diff
@@ -0,0 +1,137 @@
...
+    /** @var TwigRenderer|MockObject */
+    private $twigRenderer;
+
+    /** @var Site|MockObject */
+    private $site;
+
+    /** @var SourceFileRenderer */
+    private $sourceFileRenderer;
+
+    public function testRenderWithContentBlock() : void
```
SenseException: I would prefer the name `testRenderWithContentTwigBlock`.

tests/Builder/SourceFileRendererTest.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        $date                 = new DateTimeImmutable('2018-09-01');
+        $sourceFile           = $this->createMock(SourceFile::class);
+        $sourceFileParameters = new SourceFileParameters(['controller' => 'TestController']);
+        $controllerResult     = ['test' => true];
+        $contents             = 'Test';
```
SenseException: and with `$content`, this name would fit to the one of the test method. It's easier to read.

tests/Builder/SourceFileRendererTest.php
```diff
@@ -0,0 +1,137 @@
...
+                'test' => true,
+            ]);
+
+        $this->sourceFileRenderer->render(
+            $sourceFile,
+            $contents
+        );
+    }
+
+    public function testRenderWithoutContentBlock() : void
```
SenseException: I would prefer the name `testRenderWithoutContentTwigBlock`

tests/Builder/SourceFileRendererTest.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        $date                 = new DateTimeImmutable('2018-09-01');
+        $sourceFile           = $this->createMock(SourceFile::class);
+        $sourceFileParameters = new SourceFileParameters(['controller' => 'TestController']);
+        $controllerResult     = ['test' => true];
+        $contents             = '{% block content %}Testing{% endblock %}';
```
SenseException: And with `$content`, this name would fit to the one of the test method. It's easier to read.

tests/Builder/SourceFileRepositoryTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /** @var SourceFileRepository */
+    private $sourceFileRepository;
+
+    public function testGetFiles() : void
+    {
+        $buildDir = '';
+        $inPath   = 'source/blog';
+
+        $files = $this->sourceFileRepository->getFiles(
```
SenseException: If this test has more than one file in the return values, it is not strict enough by testing only one of the results. We should use fixtures here to control the amount of files that are returned.

tests/Builder/SourceFileRepositoryTest.php
```diff
@@ -0,0 +1,61 @@
...
+            $buildDir,
+            $inPath
```
SenseException: `$inPath` is optional and we don't have a test that covers omitted second argument.

tests/Builder/SourceFileRepositoryTest.php
```diff
@@ -0,0 +1,61 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Website\Tests\Builder;
+
+use Doctrine\Website\Builder\SourceFile;
+use Doctrine\Website\Builder\SourceFileRepository;
+use PHPUnit\Framework\TestCase;
+use function usort;
+
+class SourceFileRepositoryTest extends TestCase
```
SenseException: [Scrutinizer](https://scrutinizer-ci.com/g/doctrine/doctrine-website/inspections/85686785-d93a-4eab-8364-25a203c01ff7/code-structure/class/Doctrine%5CWebsite%5CBuilder%5CSourceFileRepository?expandCoverage=1) is showing uncovered code. We need to cover that code or remove it, in case it is unreachable.

tests/Sitemap/SitemapPageRepositoryTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Tests\Sitemap;
+
+use Doctrine\Website\Sitemap\SitemapPageRepository;
+use PHPUnit\Framework\TestCase;
+use function date;
+
+class SitemapPageRepositoryTest extends TestCase
```
SenseException: [Scrutinizer](https://scrutinizer-ci.com/g/doctrine/doctrine-website/inspections/85686785-d93a-4eab-8364-25a203c01ff7/code-structure/class/Doctrine%5CWebsite%5CSitemap%5CSitemapPageRepository?expandCoverage=1) is showing uncovered code. We need to cover that code or remove it, in case it is unreachable.

lib/Sitemap/SitemapPageRepository.php
```diff
@@ -36,10 +45,7 @@ public function findAll() : array
      */
     private function getUrlsFromFiles(string $path, string $extension = 'html') : array
```
SenseException: We actually don't need a second argument here. It is unused code.

tests/Sitemap/SitemapPageTest.php
```diff
@@ -0,0 +1,42 @@
...
+namespace Doctrine\Website\Tests\Sitemap;
+
+use DateTimeImmutable;
+use Doctrine\Website\Sitemap\SitemapPage;
+use PHPUnit\Framework\TestCase;
+
+class SitemapPageTest extends TestCase
+{
+    /** @var string */
+    private $url;
```
SenseException: This one is not needed for your tests.

lib/Sitemap/SitemapPageRepository.php
```diff
@@ -50,13 +56,29 @@ private function getUrlsFromFiles(string $path, string $extension = 'html') : ar
             RecursiveDirectoryIterator::SKIP_DOTS | RecursiveDirectoryIterator::CURRENT_AS_PATHNAME
         );
 
+        $it = new RecursiveIteratorIterator($it);
+
+        $sortedIterator = new class($it) extends SplHeap {
```
Ocramius: What's this contraption for? O_o

tests/Sitemap/SitemapPageTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function testGetDate() : void
+    {
+        self::assertSame($this->date, $this->sitemapPage->getDate());
+    }
+
+    protected function setUp() : void
+    {
+        $this->url  = '/test.html';
+        $this->date = $this->createMock(DateTimeImmutable::class);
```
SenseException: Should we use a test double for all `DateTimeImmutable` classes, where it's value doesn't matter in the tests?

lib/Sitemap/SitemapPageRepository.php
SenseException: Do we need a sorted result?

jwage: Lolll I didn't want it but I added it to get a consistent order for the test. I'm sure there is a better way but this is where I landed last night.

Ocramius: I'd be OK with just a comment explaining this :+1:

jwage: Thinking more about it this morning, I could move the sort to the test before asserting. I was getting a different order locally compared to travis. We could still get a different order of the URLs in the sitemap depending on where the site is built but we only build and push to github on the build server. What do you think?

tests/Blog/BlogPostTest.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function testGetContents() : void
+    {
+        self::assertSame('Test Content', $this->blogPost->getContents());
+    }
+
+    public function testGetDate() : void
+    {
+        self::assertSame($this->date, $this->blogPost->getDate());
```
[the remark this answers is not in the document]
jwage: It would fail because of the type hints, no? or I am not understanding what you mean.

tests/Builder/SourceFileRendererTest.php
jwage: @SenseException I don't quite understand, can you elaborate?

tests/Sitemap/SitemapPageRepositoryTest.php
jwage: I added those checks because phpstan complains otherwise. It never happens in the real scenario.

jwage: I am making some changes so it can be removed.

tests/Blog/BlogPostTest.php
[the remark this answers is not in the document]
SenseException: When someone is changing of types from `DateTimeImmutable` to `DateTime` in a PR for this class, that this test would fail.
But I guess that's what reviews are for.

tests/Builder/SourceFileRendererTest.php
SenseException: I had problems to understand at first what these tests are for. I read into the code and understood it after I looked into it. The naming of the test method and this variable I suggested here would have made its intention easier to understand (for me).

And because the naming in the test method is singular, I suggested to use singular here too. Otherwhise it would be similar like this code example:

``` php
public function findProduct()
{
    // ... code code code
    return $products;
}
```

lib/Sitemap/SitemapPageRepository.php
SenseException: If you use custom files for your test cases, you can be (pretty) sure about the order in your tests by the file names you choose. If we don't want empty test files in our repository, we can introduce mikey179/vfsStream to replace the filesystem in the tests.

If you want to keep this logic, try `SplMaxHeap` or `SplMinHeap` for the sorting instead of your custom heap. Maybe one of those do the trick for you too.

PR #216, opened 2018-08-30
Ensure we have integrity and crossorigin params on all our css and js.

### LINKED ISSUE #151: Missing integrity + crossorigin for some subresources
There are some CSS/JS linked with integrity and crossorigin attributes already, but some are missing them. :|

https://caniuse.com/#feat=subresource-integrity

jwage:
fixes #151

### DIFF
--- a/lib/Assets/AssetIntegrityGenerator.php
+++ b/lib/Assets/AssetIntegrityGenerator.php
@@ -1,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Assets;
+
+use function assert;
+use function base64_encode;
+use function file_get_contents;
+use function hash;
+use function realpath;
+
+class AssetIntegrityGenerator
+{
+    /** @var string */
+    private $sourcePath;
+
+    /** @var string[] */
+    private $cache = [];
+
+    public function __construct(string $sourcePath)
+    {
+        $this->sourcePath = $sourcePath;
+    }
+
+    public function getAssetIntegrity(string $path) : string
+    {
+        if (! isset($this->cache[$path])) {
+            $contents = $this->getFileContents($path);
+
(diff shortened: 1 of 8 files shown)


templates/layouts/layout.html.twig
```diff
@@ -257,7 +312,11 @@
             }
         </script>
 
-        <script type="text/javascript" src="https://translate.google.com/translate_a/element.js?cb=googleTranslateElementInit" async></script>
+        <script
+            type="text/javascript"
+            src="https://translate.google.com/translate_a/element.js?cb=googleTranslateElementInit"
```
jwage: When I try to add the integrity and crossorigin params on this asset, it fails with a CORS error about AllowedOrigin.

Majkl578: It's probably not a good idea to add integrity for non-versioned resource anyway. Ad crossorigin, isn't it because the script needs cookies?

lib/Assets/AssetIntegrityGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+            $this->cache[$path] = $this->buildAssetIntegrityString($contents);
+        }
+
+        return $this->cache[$path];
+    }
+
+    private function getFileContents(string $path) : string
+    {
+        $assetPath = realpath($this->sourcePath . '/' . $path);
+        assert(is_string($assetPath));
```
Majkl578: I'd prefer `$assetPath !== false` instead.

lib/Assets/AssetIntegrityGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+
+        $contents = file_get_contents($assetPath);
+        assert(is_string($contents));
```
Majkl578: dtto

PR #218, opened 2018-08-30
Allow each project to include a doctrine-project.json file in the roo…

### LINKED ISSUE #104: Allow each project to provide a project.json in the root of the project
This way each project can provide the data for the website instead of it all being in the projects.yml in doctrine/doctrine-website.  See conversation in #101 for context.

jwage:
fixes #104

### DIFF
--- a/lib/Projects/ProjectJsonReader.php
+++ b/lib/Projects/ProjectJsonReader.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Projects;
+
+use InvalidArgumentException;
+use function assert;
+use function file_exists;
+use function file_get_contents;
+use function is_string;
+use function json_decode;
+use function sprintf;
+
+class ProjectJsonReader
+{
+    private const JSON_FILENAME = 'doctrine-project.json';
+
+    /** @var string */
+    private $projectsPath;
+
+    public function __construct(string $projectsPath)
+    {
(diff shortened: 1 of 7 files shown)


lib/Projects/ProjectJsonReader.php
```diff
@@ -0,0 +1,54 @@
...
+            return null;
+        }
+
+        $projectJsonString = file_get_contents($projectJsonPath);
+
+        assert(is_string($projectJsonString));
+
+        $projectJson = json_decode($projectJsonString, true);
+
+        if ($projectJson === false || $projectJson === null) {
```
stof: JSON decoding can also return a string, a number or a boolean, which would break the return type. You should use `!is_array` instead to account for any invalid data.

And if you account only for invalid JSON, the `false` check is useless (`false` can be returned only by decoding the `false` JSON string) and `null` is the wrong check (as it might also be a valid `null` JSON string)

jwage: @stof I put a fix for this here https://github.com/doctrine/doctrine-website/pull/220

Majkl578: Preferably the check for error should use `json_last_error() !== JSON_ERROR_NONE`, or `JSON_THROW_ON_ERROR` flag as of 7.3.

jwage: Added that in https://github.com/doctrine/doctrine-website/pull/220

PR #219, opened 2018-08-30
Publish and refine our deprecation policy.

jwage:
screen shot 2018-09-01 at 7 29 45 pm

screen shot 2018-09-01 at 7 29 51 pm

### DIFF
--- a/source/policies/deprecation.rst
+++ b/source/policies/deprecation.rst
@@ -1,0 +1,165 @@
+---
+layout: default
+title: Deprecation Policy
+menuSlug: development
+permalink: /policies/deprecation.html
+---
+
+Deprecation Policy
+==================
+
+The Doctrine team will adhere to this deprecation policy whenever it is reasonably possible.
+
+Using ``@deprecated`` and ``PHPStan Deprecation Rules``
+-------------------------------------------------------
+
+The primary and preferred method for deprecating functionality within Doctrine is by using
+the ``@deprecated`` annotation. To have usages of functionality in your dependencies analyzed
+and reported, you must use a tool like `PHPStan Deprecation Rules`_ or `Pslam Static Analysis`_.
+
+Here is an example of how this would be used in Doctrine:
+
+.. code-block:: php
+
+    namespace Doctrine\ORM;
+
+    class EntityManager
+    {
+        // ...
+
+        /**
+         * @deprecated
+         */
+        public function getSomething() : int
+        {
+            // ...
+        }
+    }
+
+Then, if your code directly depends on ``doctrine/orm`` which provides the
+``Doctrine\ORM\EntityManager`` class, you will get a warning about it when you run
+`PHPStan Deprecation Rules`_ or `Pslam Static Analysis`_. Here is an example:
+
+.. code-block:: php
+
+    namespace App;
+
+    use App\Entities\User;
+    use Doctrine\ORM\EntityManager;
+
+    class UserRegistration
+    {
+        /** @var EntityManager */
+        private $entityManager;
+
+        // ...
+
+        public function create(array $userData) : User
+        {
+            $something = $this->entityManager->getSomething();
+
+            // ...
+        }
+    }
+
+Using ``trigger_error()`` and ``phpunit-bridge``
+------------------------------------------------
+
+In some cases, you may need to conditionally deprecate functionality. In these cases, it is required
+to use the ``trigger_error()`` function and a tool like `PHPUnit Bridge`_ which implements a custom
+error handler for you to use to collect and report these warnings. **This method is not recommended
+and Doctrine will try to avoid using it if possible.**
+
+If you do not wish to use `PHPUnit Bridge`_ you can implement your own custom error handler using
+the `set_error_handler`_ function.
+
+.. note::
+
+    In order for this strategy to work, you have to configure your `error_reporting`_ properly
+    in both your development and production environments. You don't want errors of type ``E_USER_DEPRECATED``
+    being reported from PHP.
+
+Here is an example of how this would be used in Doctrine:
+
+.. code-block:: php
+
+    namespace Doctrine\ORM;
+
+    class EntityManager
+    {
+        // ...
+
+        public function getAnotherThing() : int
+        {
+            if ($this->featureFlag) {
+                trigger_error('This feature has been disabled.', E_USER_DEPRECATED);
+            }
+
+            // ...
+        }
+    }
+
+Now in order to get those warnings reported to you in your automation, you need to use a tool like `PHPUnit Bridge`_.
+
+Tools
+-----
+
+There are a few tools out there that you can use to integrate deprecation warnings into your
+automation.
+
+PHPStan Deprecation Rules
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Doctrine uses and recommends PHPStan_ for reporting usages of deprecated functionality in your code.
+
+.. code-block:: console
+
+    $ composer require --dev phpstan/phpstan
+    $ composer require --dev phpstan/phpstan-deprecation-rules
+
+You will need to setup a configuration file in the root of project named ``phpstan.neon.dist``
+
+.. code-block::
+
+    includes:
+        - vendor/phpstan/phpstan-phpunit/extension.neon
+        - vendor/phpstan/phpstan-strict-rules/rules.neon
+
+    parameters:
+        level: 0
+        paths:
+          - lib
+          - tests
+
+Now you can run the ``vendor/bin/phpstan`` command:
+
+.. code-block:: console
+
+    $ ./vendor/bin/phpstan analyze
+
+Pslam Static Analysis
+~~~~~~~~~~~~~~~~~~~~~
+
+Psalm_ is a static analysis tool for finding errors in PHP applications, built on top of `PHP Parser`_.
+
+It's able to find a large number of issues, but it can also be configured to only care about a small subset of those.
+
+.. code-block:: console
+
+    $ composer require vimeo/psalm
+
+PHPUnit Bridge
+~~~~~~~~~~~~~~
+
+The `PHPUnit Bridge`_ provides utilities to report legacy tests and usage of deprecated code.
+
+.. code-block:: console
+
+    $ composer require --dev "symfony/phpunit-bridge:*"
+
+.. _Pslam: https://github.com/vimeo/psalm
+.. _PHPStan: https://github.com/phpstan/phpstan
+.. _PHP Parser: https://github.com/nikic/php-parser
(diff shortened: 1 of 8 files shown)


source/deprecation-policy.rst
```diff
@@ -0,0 +1,145 @@
...
+            // ...
+        }
+    }
+
+Now in order to get those warnings reported to you in your automation, you need to use a tool like `PHPUnit Bridge`_.
+
+Tools
+-----
+
+There are a few tools out there that you can use to integrate deprecation warnings in to your
```
SenseException: "in to" -> "into"?

source/deprecation-policy.rst
```diff
@@ -0,0 +1,145 @@
...
+automation.
+
+PHPStan Deprecation Rules
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Doctrine uses and recommends PHPStan_ for reporting usages of deprecated functionality in your code.
```
SenseException: &#96;PHPStan_&#96;

It is usable without backticks on Github, but is this the same case for our documentation?

source/deprecation-policy.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. code-block:: console
+
+    $ composer require --dev phpstan/phpstan
+    $ composer require --dev phpstan/phpstan-deprecation-rules
+
+Pslam Static Analysis
+~~~~~~~~~~~~~~~~~~~~~
+
+Psalm_ is a static analysis tool for finding errors in PHP applications, built on top of `PHP Parser`_.
```
SenseException: Do we need backticks for Psalm in our Doctrine website or is it handled without problems in our documentation?

source/deprecation-policy.rst
jwage: Yes, it works. I was building it locally as I wrote  it.

source/deprecation-policy.rst
jwage: It works with our rst parser.

source/deprecation-policy.rst
```diff
@@ -0,0 +1,145 @@
...
+
+PHPStan Deprecation Rules
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Doctrine uses and recommends PHPStan_ for reporting usages of deprecated functionality in your code.
+
+.. code-block:: console
+
+    $ composer require --dev phpstan/phpstan
+    $ composer require --dev phpstan/phpstan-deprecation-rules
```
lcobucci: Should we provide a sample config file? The extension doesn't get automatically enabled by just installing it via composer, one needs to append the extension file to the configuration, like https://github.com/lcobucci/kafka/blob/1b1284020f20018e2d54db2d97ee30e91dff7708/phpstan.neon.dist#L5

templates/layouts/layout.html.twig
```diff
@@ -131,12 +131,14 @@
...
                         </li>
                         <li class="nav-item dropdown{% if menuSlug == 'contribute' %} active{% endif %}">
-                            <a class="nav-link dropdown-toggle" href="{{ site.url }}/contribute/" id="navbarContributeDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Contribute</a>
+                            <a class="nav-link dropdown-toggle" href="{{ site.url }}/contribute/" id="navbarDevelopmentDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Development</a>
 
                             <div class="dropdown-menu" aria-labelledby="navbarContributeDropdown">
                                 <a class="dropdown-item" href="{{ site.url }}/contribute/">Contributor Workflow</a>
                                 <a class="dropdown-item" href="{{ site.url }}/contribute/maintainer/">Maintainer Workflow</a>
                                 <a class="dropdown-item" href="{{ site.url }}/contribute/website/">Contribute to Website</a>
+                                <a class="dropdown-item" href="{{ site.url }}/deprecation-policy/">Deprecation Policy</a>
```
lcobucci: Should we have a link with all of our policies? Otherwise the menu might get a bit big as time passes by

Ocramius:
@nicolasgrekas let's skim over this together today 👍

Ocramius:
@nicolas-grekas let's skim over this together today, if you got time 👌

source/deprecation-policy.rst
```diff
@@ -0,0 +1,145 @@
...
+            // ...
+        }
+    }
+
+Using ``trigger_error()`` and ``phpunit-bridge``
+------------------------------------------------
+
+In some cases, you may need to conditionally deprecate functionality. In these cases, it is required
+to use the ``trigger_error()`` function and a tool like `PHPUnit Bridge`_ which implements a custom
+error handler for you to use to collect and report these warnings. This method is not recommended
```
lcobucci: > This method is not recommended and Doctrine will try to avoid using it if possible.

What do you think about making this bold? Just to emphasise a bit more :+1:

templates/layouts/layout.html.twig
jwage: Sure, I can go ahead and setup that structure to support other policies like releases, security, etc.

lcobucci: Awesome, thanks!

Ocramius: Note: releases is something I'm currently working on :-)

source/deprecation-policy.rst
```diff
@@ -0,0 +1,145 @@
...
+
+            // ...
+        }
+    }
+
+Using ``trigger_error()`` and ``phpunit-bridge``
+------------------------------------------------
+
+In some cases, you may need to conditionally deprecate functionality. In these cases, it is required
+to use the ``trigger_error()`` function and a tool like `PHPUnit Bridge`_ which implements a custom
```
Ocramius: Discussed this with @nicolas-grekas in person: the `PHPUnit_Bridge` is not needed. With out of the box PHPUnit, this is the behavior:

Passes:

```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

final class SomeTest extends \PHPUnit\Framework\TestCase
{
    public function testFoo() : void
    {
        @trigger_error(E_USER_DEPRECATED, 'potato');
        self::assertTrue(true);
    }
}
```
```sh
./vendor/bin/phpunit SomeTest.php 
PHPUnit 7.3.2 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)

Time: 60 ms, Memory: 4.00MB

OK (1 test, 1 assertion)
```


Fails:

```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

final class SomeTest extends \PHPUnit\Framework\TestCase
{
    public function testFoo() : void
    {
        trigger_error(E_USER_DEPRECATED, 'potato');
        self::assertTrue(true);
    }
}
```
```sh
./vendor/bin/phpunit SomeTest.php 
PHPUnit 7.3.2 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 60 ms, Memory: 4.00MB

There was 1 error:

1) SomeTest::testFoo
trigger_error() expects parameter 2 to be integer, string given

/home/ocramius/foo/SomeTest.php:9

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

Note that this is because PHPUnit by default checks the error level before turning it into an error, while following still leads to failures:

```php
<?php

set_error_handler(function () {
    echo 'I died :-(';
    die;
});

@trigger_error(E_USER_DEPRECATED, 'foo');
```

Produces:

```
I died :-(
```

This means that users that just use `set_error_handler()` (no parameters, no filtering, etc) will get a BC break.

As per discussion with @nicolas-grekas (where this surfaced as well), this is a decision to be taken, because some users may be doing it "wrong", but effectively I've already seen codebases where zero-warnings-or-notices-ever are accepted, and those codebases will be damaged by using `@trigger_error()`.

In my personal opinion, we should keep things towards `@deprecated`.

nicolas-grekas:
I definitely want to have a look here, and maybe write down Symfony's practices on the topic
For reference, here is Drupal document:
https://www.drupal.org/core/deprecation

source/deprecation-policy.rst
lcobucci: @Ocramius I think you inverted the order of arguments in `trigger_error() ` calls.

> In my personal opinion, we should keep things towards `@deprecated`. 

I agree, we should strive for using `@deprecated` as much as possible.

nicolas-grekas:
Ok read. Not adding the `@` in front of `trigger_error()` would be a mistake IMHO, but you'll figure out later I suppose. I don't have more specific comments on the other parts of the document.

source/deprecation-policy.rst
Ocramius: Indeed, the arguments were swapped :-\

source/policies/deprecation.rst
```diff
@@ -0,0 +1,165 @@
...
+==================
+
+The Doctrine team will adhere to this deprecation policy whenever it is reasonably possible.
+
+Using ``@deprecated`` and ``PHPStan Deprecation Rules``
+-------------------------------------------------------
+
+The primary and preferred method for deprecating functionality within Doctrine is by using
+the ``@deprecated`` annotation. To have usages of functionality in your dependencies analyzed
+and reported, you must use a tool like `PHPStan Deprecation Rules`_ or `Pslam Static Analysis`_.
```
Majkl578: typo: Psalm

source/policies/deprecation.rst
```diff
@@ -0,0 +1,165 @@
...
+PHPUnit Bridge
+~~~~~~~~~~~~~~
+
+The `PHPUnit Bridge`_ provides utilities to report legacy tests and usage of deprecated code.
+
+.. code-block:: console
+
+    $ composer require --dev "symfony/phpunit-bridge:*"
+
+.. _Pslam: https://github.com/vimeo/psalm
```
Majkl578: typo: Psalm

source/policies/deprecation.rst
```diff
@@ -0,0 +1,165 @@
...
+          - lib
+          - tests
+
+Now you can run the ``vendor/bin/phpstan`` command:
+
+.. code-block:: console
+
+    $ ./vendor/bin/phpstan analyze
+
+Pslam Static Analysis
```
Majkl578: typo: Psalm

source/policies/deprecation.rst
```diff
@@ -0,0 +1,165 @@
...
+         */
+        public function getSomething() : int
+        {
+            // ...
+        }
+    }
+
+Then, if your code directly depends on ``doctrine/orm`` which provides the
+``Doctrine\ORM\EntityManager`` class, you will get a warning about it when you run
+`PHPStan Deprecation Rules`_ or `Pslam Static Analysis`_. Here is an example:
```
Majkl578: typo: Psalm

PR #220, opened 2018-08-31
Read composer json data to try and get data for the website.

jwage:


lib/Projects/ProjectFactory.php
```diff
@@ -27,10 +28,8 @@ public function create(array $project) : Project
 
         $projectJson = $this->projectJsonReader->read($project['repositoryName']);
 
-        if ($projectJson !== null) {
-            $project = $project + $projectJson;
-        }
+        $projectData = array_merge($projectJson, $project);
```
stof: as you deal with associative arrays, why not `array_replace` instead ?

jwage: Done.

lib/Projects/ProjectJsonReader.php
```diff
@@ -27,28 +32,70 @@ public function __construct(string $projectsPath)
...
+     * @return mixed[]
+     */
+    private function readComposerData(string $repositoryName) : array
     {
-        $projectJsonPath = $this->projectsPath . '/' . $repositoryName . '/' . self::JSON_FILENAME;
+        $data = $this->readJsonFile($repositoryName, self::COMPOSER_JSON_FILE_NAME);
 
-        if (! file_exists($projectJsonPath)) {
-            return null;
+        $composerData = [];
```
Majkl578: This looks like a decent candidate for a(n immutable) DTO. :)

PR #224, opened 2018-09-01
Expand "Who uses Doctrine" section

Majkl578:
Add some more projects that use Doctrine.

PR #225, opened 2018-09-01
Make the who uses doctrine section configurable and add a functional …

jwage:
Still time to add @asgrim :)

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -16,3 +16,47 @@ parameters:
         are the Object Relational Mapper (ORM) and the Database Abstraction
         Layer (DBAL) it is built upon.
     doctrine.website.google_analytics_tracking_id: ''
+
+    doctrine.website.who_uses_doctrine:
+        -
+            name: Symfony
+            url: https://symfony.com/
+        -
+            name: Drupal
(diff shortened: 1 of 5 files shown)


Majkl578:
The projects I added in #224 were deliberately picked pseudo-randomly (popularity, downloads) from reverse dependencies on Packagist. Also I didn't want to (and would like not to) include any non-OSS projects here on homepage - we may create separate page for those if desired.

I don't know what project(s) @asgrim meant though.

jwage:
I was just joking about @asgrim 😁

config/config.yml
```diff
@@ -16,3 +16,47 @@ parameters:
...
+            name: PHPUnit
+            url: https://phpunit.de/
+        -
+            name: Prophecy
+            url: https://github.com/phpspec/prophecy
+        -
+            name: phpbench
+            url: https://github.com/phpbench/phpbench
+        -
+            name: Typo3
```
Majkl578: Can you please change this to `TYPO3`? I got this one wrong.

jwage: Done.

PR #226, opened 2018-09-01
Fix blog index style so it looks like it did before migrating away fr…

jwage:
Before

screen shot 2018-09-01 at 3 01 02 pm

After

screen shot 2018-09-01 at 3 01 22 pm

### DIFF
--- a/lib/Blog/BlogPostRepository.php
+++ b/lib/Blog/BlogPostRepository.php
@@ -6,8 +6,10 @@ namespace Doctrine\Website\Blog;
 
 use Doctrine\Website\Builder\SourceFile;
 use Doctrine\Website\Builder\SourceFileRepository;
+use InvalidArgumentException;
 use function array_map;
 use function array_reverse;
+use function array_slice;
 use function usort;
 
 class BlogPostRepository
@@ -30,6 +32,20 @@ class BlogPostRepository
             $sourceFile->getContents(),
             $sourceFile->getDate()
         );
+    }
+
+    /**
+     * @return BlogPost[]
+     */
+    public function findPaginated(int $page = 1, int $perPage = 10) : array
+    {
+        if ($page < 1 || $perPage < 1) {
+            throw new InvalidArgumentException('Pagination parameters must be positive.');
+        }
+
+        $offset = ($page - 1) * $perPage;
+
+        return array_slice($this->findAll(), $offset, $perPage);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


lib/Controllers/BlogController.php
```diff
@@ -19,6 +20,13 @@ public function __construct(BlogPostRepository $blogPostRepository)
     }
 
     public function index(SourceFile $sourceFile) : ControllerResult
+    {
+        return new ControllerResult([
+            'blogPosts' => array_slice($this->blogPostRepository->findAll(), 0, 10),
```
SenseException: We should put this slicing into the repository with an limit argument.

jwage: Done.

templates/layouts/blog-post.html.twig
```diff
@@ -43,7 +43,7 @@
         </header>
 
         <p class="lead">
-            Posted on {{ blogPost.date|date('Y-m-d') }}
+            Posted on {{ blogPost.date|date('F j, Y') }}
```
Ocramius: Same here: need `<date/>` around it

source/archive.html
```diff
@@ -0,0 +1,23 @@
...
+title: Blog Archive
+menuSlug: blog
+permalink: /blog/archive.html
+controller: ['Doctrine\Website\Controllers\BlogController', 'archive']
+---
+
+<h1>Blog Archive</h1>
+
+{% for blogPost in blogPosts %}
+    {{ blogPost.date|date('Y-m-d') }} &nbsp; <a href="{{ blogPost.url }}">{{ blogPost.title }}</a>
```
Ocramius: Let's use `<date/>` around this

jwage: 👍Changed

lib/Blog/BlogPostRepository.php
```diff
@@ -32,6 +33,16 @@ public function find(SourceFile $sourceFile) : BlogPost
         );
     }
 
+    /**
+     * @return BlogPost[]
+     */
+    public function findPaginated(int $page = 1, int $perPage = 10) : array
```
SenseException: How should this behave on values `< 1`? Throwing an exception? Can we use `assert` here?

jwage: Hmm, good question. I am not sure. Do you think we should do something?

SenseException: `array_slice` allows negative values for offset and length, but that's an implementation detail, that is not known to the code invoking `findPaginated()`. Are negative numbers allowed by the repository method? If not, then we should check them.

jwage: I get what you are saying. Do you think it is worth it in such a simple application where we are the ones calling findPaginated in 2 places?

Ocramius: Yes

PR #246, opened 2018-09-14
Enhance projects list

jwage:
screen shot 2018-09-13 at 8 06 17 pm

### DIFF
--- a/lib/Projects/ProjectDataReader.php
+++ b/lib/Projects/ProjectDataReader.php
@@ -30,13 +30,21 @@ class ProjectDataReader
     /** @var mixed[] */
     private $projectsData;
 
+    /** @var mixed[] */
+    private $projectIntegrationTypes;
+
     /**
      * @param mixed[] $projectsData
+     * @param mixed[] $projectIntegrationTypes
      */
-    public function __construct(string $projectsPath, array $projectsData)
-    {
-        $this->projectsPath = $projectsPath;
-        $this->projectsData = $projectsData;
+    public function __construct(
+        string $projectsPath,
+        array $projectsData,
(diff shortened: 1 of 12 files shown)


SenseException:
The build of Depndency=low on travis is green, but skips some tests: "This test requires ./bin/console build-website to have been ran."

lib/Projects/ProjectDataReader.php
```diff
@@ -30,26 +30,55 @@ class ProjectDataReader
...
+        $this->projectIntegrationTypes = $projectIntegrationTypes;
     }
 
     /**
      * @return mixed[]
      */
     public function read(string $repositoryName) : array
     {
-        return array_replace(
+        $projectData = array_replace(
```
SenseException: Would it make sense to create a VO for the project data?

jwage: This is the array of data used to populate that Project model/vo.

jwage:
The full build only runs on the coverage stage. It builds the full website and then runs the functional tests.

lib/Projects/ProjectDataReader.php
SenseException: I know, but it isn't unusual that a VO gets other VOs injected. Project data itself is now used by two other VOs. We can have invalid (well, theoretically, because there are array key checks) arrays that should fail earlier by using a VO and may reduce later checks with `if` or `??`.

I made this a comment instead of a change request, because it isn't currently important for an approval. I just see a point where we can start to introduce the suggested VO.

jwage: I will look at making that improvement in another PR.

PR #247, opened 2018-09-19
Use Doctrine Skeleton Mapper for the object mapper layer.

jwage:
Playing around with using https://github.com/doctrine/skeleton-mapper for the object mapper layer. This makes it so we can easily manage models where the data is provided from arrays of data provided by container parameters.


SenseException:
A lot of files have changed. I'll tests the branch itself too.

lib/DataSources/TeamMembers.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * @return mixed[][]
+     */
+    public function getData() : array
+    {
+        $teamMembers = [];
+
+        foreach ($this->teamMembers as $key => $teamMember) {
```
SenseException: The key is not used here. I kind of expect that the static code analysis would find something like that.

lib/Model/SitemapPage.php
```diff
@@ -0,0 +1,50 @@
...
+        $metadata->setIdentifier(['url']);
+    }
+
+    /**
+     * @param mixed[] $sitemapPage
+     */
+    public function hydrate(array $sitemapPage, ObjectManagerInterface $objectManager) : void
+    {
+        $this->url  = $sitemapPage['url'] ?? '';
+        $this->date = $sitemapPage['date'] ?? '';
```
SenseException: `DateTimeImmutable`. Don't put a string into the date property.

lib/Model/BlogPost.php
```diff
@@ -42,6 +46,24 @@ public function __construct(
...
 
+    public static function loadMetadata(ClassMetadataInterface $metadata) : void
+    {
+        $metadata->setIdentifier(['url']);
+    }
+
+    /**
+     * @param mixed[] $project
+     */
+    public function hydrate(array $project, ObjectManagerInterface $objectManager) : void
```
SenseException: Types for the properties should be handled here or we can get an invalid object state. Too bad `hydrate()` doesn't expect an Exception.

lib/Model/Contributor.php
```diff
@@ -0,0 +1,91 @@
...
+
+    public static function loadMetadata(ClassMetadataInterface $metadata) : void
+    {
+        $metadata->setIdentifier(['github']);
+    }
+
+    /**
+     * @param mixed[] $contributor
+     */
+    public function hydrate(array $contributor, ObjectManagerInterface $objectManager) : void
```
SenseException: Types for the properties should be handled in this method or we can get an invalid object state.

lib/Model/DoctrineUser.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public static function loadMetadata(ClassMetadataInterface $metadata) : void
+    {
+        $metadata->setIdentifier(['name']);
+    }
+
+    /**
+     * @param mixed[] $doctrineUser
+     */
+    public function hydrate(array $doctrineUser, ObjectManagerInterface $objectManager) : void
```
SenseException: Types for the properties should be handled in this method or we can get an invalid object state.

lib/Model/ProjectContributor.php
```diff
@@ -0,0 +1,97 @@
...
+
+    public static function loadMetadata(ClassMetadataInterface $metadata) : void
+    {
+        $metadata->setIdentifier(['projectSlug', 'github']);
+    }
+
+    /**
+     * @param mixed[] $projectContributor
+     */
+    public function hydrate(array $projectContributor, ObjectManagerInterface $objectManager) : void
```
SenseException: Types for the properties should be handled in this method or we can get an invalid object state.

lib/Model/SitemapPage.php
```diff
@@ -0,0 +1,50 @@
...
+
+    public static function loadMetadata(ClassMetadataInterface $metadata) : void
+    {
+        $metadata->setIdentifier(['url']);
+    }
+
+    /**
+     * @param mixed[] $sitemapPage
+     */
+    public function hydrate(array $sitemapPage, ObjectManagerInterface $objectManager) : void
```
SenseException: Types for the properties should be handled in this method or we can get an invalid object state.

lib/Model/TeamMember.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public static function loadMetadata(ClassMetadataInterface $metadata) : void
+    {
+        $metadata->setIdentifier(['github']);
+    }
+
+    /**
+     * @param mixed[] $teamMember
+     */
+    public function hydrate(array $teamMember, ObjectManagerInterface $objectManager) : void
```
SenseException: Types for the properties should be handled in this method or we can get an invalid object state.

tests/DataSources/BlogPostsTest.php
```diff
@@ -0,0 +1,95 @@
...
+        $files = [$sourceFile1, $sourceFile2];
+
+        $this->sourceFileRepository->expects(self::once())
+            ->method('getFiles')
+            ->with('', 'source/blog')
+            ->willReturn($files);
+
+        $blogPostsData = $this->blogPosts->getData();
+
+        usort($blogPostsData, new Sorter(['date' => 'desc']));
```
SenseException: Please bring the expected result into the expected order instead of using/introducing a test-unrelated logic.

composer.json
```diff
@@ -13,9 +13,13 @@
     "require": {
         "php": "^7.1",
         "algolia/algoliasearch-client-php": "^1.27",
+        "cache/doctrine-adapter": "^1.0",
         "doctrine/inflector": "^1.3",
         "doctrine/rst-parser": "dev-master",
+        "doctrine/skeleton-mapper": "dev-master",
         "erusev/parsedown": "^1.7",
+        "knplabs/github-api": "^2.10",
+        "php-http/guzzle6-adapter": "^1.1",
```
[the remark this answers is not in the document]
jwage: It was required by the github client.

Unrelated but this is why the team data was empty for you. You need to configure the github api key. I have it catching the exception right now. I need to remove that and fix it so I can configure the api key on travis so the tests don't fail.

config/services.xml
```diff
@@ -33,10 +33,27 @@
...
+                    </service>
+                </argument>
+            </call>
+        </service>
         <service id="Highlight\Highlighter" autowire="true" />
         <service id="Parsedown" class="Parsedown" autowire="true" />
         <service id="Symfony\Component\Console\Application" autowire="true" />
         <service id="Symfony\Component\Filesystem\Filesystem" autowire="true" />
+        <service id="Doctrine\Common\Cache\FilesystemCache" autowire="false">
+            <argument>%doctrine.website.cache_dir%</argument>
```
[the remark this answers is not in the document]
jwage: I think I didn't use auto wiring here because it is not a class I control so the argument names are not very specific.

lib/DataSource/DataSourceObjectDataRepository.php
```diff
@@ -0,0 +1,142 @@
...
+
+namespace Doctrine\Website\DataSource;
+
+use Doctrine\SkeletonMapper\DataRepository\BasicObjectDataRepository;
+use Doctrine\SkeletonMapper\ObjectManagerInterface;
+use function array_slice;
+use function in_array;
+use function usort;
+
+class DataSourceObjectDataRepository extends BasicObjectDataRepository
```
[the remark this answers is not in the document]
jwage: In the skeleton mapper you have two interfaces. The first is how you provide the raw data to the mapper called ObjectDataRepository. The other is ObjectRepository which you are familiar with from the other Doctrine object mapper implementations.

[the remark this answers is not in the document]
jwage: This is maybe a signal the naming could be better, but that is how it is today in doctrine/skeleton mapper.

lib/DataSource/Sorter.php
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Doctrine\Website\DataSource;
+
+use RuntimeException;
+use function count;
+use function is_string;
+use function sprintf;
+use function strtolower;
+
+class Sorter
```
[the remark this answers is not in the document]
jwage: I will add tests for everything in this PR before we merge.

config/services.xml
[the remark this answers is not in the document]
jwage: The FilesystemCache constructor looks like:

    public function __construct($directory, $extension = self::EXTENSION, $umask = 0002)

So I would have to bind a pretty ambiguous argument variable name. It felt better to be explicit here. Thoughts?

jwage:
I renamed `DataSource::getData()` to `DataSource::getSourceRows()` but I am not sure about it either. It is better, but I am still not 100% happy with it.

The tests fail now because I removed the catching of the exception when the GitHub API fails. I need to figure out how to use TravisCI environment variables so that I can set the github API key so it can retrieve all the team/contributors data properly.

tests/DataSources/BlogPostsTest.php
SenseException: The tested method shouldn't be sorting at all and the mock object controls in what order the result is returned, so the test shouldn't have a surprise order. A DataSource doesn't sort any values, so I think that even `array_reverse` isn't needed in `BlogPosts`. The sorting will be done by the Repository.

lib/Model/BlogPost.php
SenseException: I'm looking forward to typed properties too. The hydrator is currently the part where no wrong types should be able to reach its destined property.

lib/DataSources/ProjectContributors.php
```diff
@@ -0,0 +1,112 @@
...
+                $numDeletions = 0;
+
+                foreach ($contributor['weeks'] as $week) {
+                    $numAdditions += $week['a'];
+                    $numDeletions += $week['d'];
+                }
+
+                $teamMember = $this->teamMemberRepository->findOneByGithub($contributor['author']['login']);
+
+                $isMaintainer = $teamMember !== null
```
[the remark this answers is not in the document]
SenseException: This line uses the new variable too.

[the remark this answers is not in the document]
jwage: phpstan doesn't like it. I guess it isn't smart enough to handle this case right now.

```
1032 /data/doctrine-website (object-mapper) $ phpstan a
Note: Using configuration file /data/doctrine-website/phpstan.neon.dist.
 122/122 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ -------------------------------------------------------------------------------------
  Line   lib/DataSources/ProjectContributors.php
 ------ -------------------------------------------------------------------------------------
  68     Cannot call method isProjectMaintainer() on Doctrine\Website\Model\TeamMember|null.
 ------ -------------------------------------------------------------------------------------


 [ERROR] Found 1 error
```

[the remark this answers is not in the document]
SenseException: Yeah, I tried some combinations but it looks like phpstan doesn't understand this case.

SenseException:
I'll take a look at it soon. :+1:

PR #251, opened 2018-09-28
Read project version data automatically.

jwage:
Working on trying to automate the project versions data so we don't have to update as much with every release. Currently using the packagist API but I don't actually need to do it this way. We can just use git tag to list all the tags and then I can build the versions from that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ jobs:
         - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{.disabled,}
         - if [[ ! $(php -m | grep -si xdebug) ]]; then echo "xdebug required for coverage"; exit 1; fi
       script:
+        - ./bin/console build-website-data
         - ./bin/console build-docs --sync-git --env=test
         - ./bin/console build-website --env=test
         - ./vendor/bin/phpunit --coverage-clover clover.xml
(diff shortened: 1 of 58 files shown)


jwage:
Yes

jwage:
You need to have the website data built before you can build the website. I wanted to separate the processes so I can build the data once and then build the website multiple times much faster when testing.

jwage:
Sorry, I misunderstood your previous comment. The `clear-build-cache` command depends on the cache being populated. I can change the cache command to work in a more general way to avoid this. I have had plans to change the `clear-build-cache` for other reasons already too.

jwage:
I pushed up a change to `ClearBuildCacheCommand` to depend on a new `CacheClearer` service and I changed it to not depend on the `Project` model and the `projects.json` data.

SenseException:
I'll take a look at it. Thank you. :smiley:

jwage:
- The scrutinizer issue is a bug in scrutinizer. Nothing is wrong that needs fixing.
- I added an ini_set for memory_limit in the build-website-data command
- I refactored `ProjectDataBuilder` to improve the rating on scrutinizer.

SenseException:
As this is now, we would always get a broken build in case this isn't a false positive but a bug. This can be confusing for contributors. What benefits does Scrutinizer have for the website that can't be covered by other static analysis tools?

Majkl578:
It only reports touched code.

PR #252, opened 2018-09-28
Update readme to include additional configuration information for bui…

jwage:
…lding the website.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ jobs:
         - rm composer.lock
         - travis_retry composer update --prefer-dist --prefer-lowest
       script:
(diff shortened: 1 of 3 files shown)


Ocramius:
```
1) Doctrine\Website\Tests\FunctionalTest::testSitemap
Failed asserting that file "/home/travis/build/doctrine/doctrine-website/tests/../build-dev/sitemap.xml" exists.
```

Expected?

jwage:
@Ocramius oh hmm, this is because I added that placeholder parameter to `local.yml.dist` and the travis ci build copies that file over. We don't need to do that, let me remove that from the travis config.

PR #254, opened 2018-10-02
Coverted into scss

BhautikChudasama:


lex111:
To be honest, I see no advantages in this :-1:

BhautikChudasama:
Sorry but I only converted into scss!

Issue #253, opened 2018-10-02
Cleanup style.css


jwage:
https://github.com/doctrine/doctrine-website/blob/master/source/css/style.css

This file is a disaster. It could be cleaned up. Maybe adopt sass and/or tailwindcss.

BhautikChudasama:
Do you need to convert it to scss?

jwage:
Yes, that could be a good first step. We'd need to create a build process too.

mmollick:
Looks like the previous PR's stalled out and never implemented the build step. I'd be happy to implement a build step and pre-commit linting for the styles. 
Are there any preferences on the tools used to accomplish this (e.g. Webpack vs Gulp)? 
While were at should we also add a build step to concat & minify the Javascript as well (Webpack makes this simple)?

SenseException:
My personal choice would be Webpack. This might be useful for changes of the website's Javascript and JS libraries for bundling.

PR #255, opened 2018-10-02
cleanup style.css and create min-style.css

thiagormagalhaes:
to issue #253


source/css/style.css
```diff
@@ -1,573 +1,510 @@
...
 }
 
-h1, h2, h3, h4, h5, h6 {
-    margin-bottom: 1rem;
+h1 {
+  margin-bottom: 1rem;
 }
 
-.search-results {
-    margin-left: 20px;
```
Ocramius: Please don't change indentation everywhere according to your editor of preference

PR #260, opened 2018-10-25
Automatically generate project pages from project data.

jwage:


jwage:
Just going to add a few new unit tests before I merge.

PR #261, opened 2018-10-27
Add webpack for frontend assets

mmollick:
Introduces webpack for frontend asset compilation.
- Adds NodeJS dependencies via [Yarn](https://yarnpkg.com) (includes lock file for consistency)
- Adds commands to build assets for production, development and file change watching `npm run [build|dev|watch]`
- Adds the previous NPM command to `build-website`
  - Assets are built in `./.webpack-build` and then symlinked (non-publishable) or copied (publishable) to the site build directory
- Converts CSS to Sass (direct conversion of CSS)
- Adds new `get_webpack_asset_url` and `get_webpack_asset_integrity` Twig functions to load Webpack generated assets
- Updates website contribution guide to include new commands
- Adds [Prettier](https://prettier.io/) to pre-commit hook to enforce Javascript and Sass file formatting.

This partially completes #253, migrating to tailwind will require additional work.


SenseException:
Thank you for your PR. Could you please fix the coding styles and take a look at the failing tests?

mmollick:
It appears Scrutinizer is using an outdated version of Node but I don't see a config file to update this in. How is Scrutinizer currently configured?

I'm also unsure what the error in Travis is from, it appears it's having an issue getting data from GitHub. Has this exception been encountered before?
```
In GithubExceptionThrower.php line 87:
                   
  Bad credentials  
                   
```

jwage:
The tests currently hit the github API and it looks like it is failing now. The same api token works for me locally so I am wondering if github blocked travis or something. I am going to need to work on a change so that the functional tests don't need to hit the github api.

mmollick:
It looks like Travis might be hitting some API limits. With the most recent push it failed with the message: `You have reached GitHub hourly limit! Actual limit is: 60`. I've made the code quality changes for phpcs.

Any suggestions on how to configure Scrutinizer to use a more recent version of NodeJS would be appreciated. I'm not sure if there's a configuration file in the repo I'm missing or not.

jwage:
I think we need to add a .scrutinizer.yml and look at the Scrutinizer documentation for how to do that. You can see examples of other .scrutinizer.yml files in other Doctrine repos like https://github.com/doctrine/doctrine2/blob/master/.scrutinizer.yml

I have never used nodejs with scrutinizer though so we'll have to do some research to figure that out.

mmollick:
Thanks, I'll start with that config and work from there.

.scrutinizer.yml
```diff
@@ -0,0 +1,35 @@
...
+
+filter:
+    excluded_paths:
+        - docs
+
+build_failure_conditions:
+    - 'elements.rating(<= C).new.exists'                        # No new classes/methods with a rating of C or worse allowed
+    - 'issues.severity(>= MAJOR).new.exists'                    # New issues of major or higher severity
+    - 'project.metric_change("scrutinizer.test_coverage", < 0)' # Code Coverage decreased from previous inspection
+
```
jwage: Remove extra new line here.

package.json
```diff
@@ -0,0 +1,24 @@
+{
```
jwage: 4 spaces instead of 2 to match the other .json files in the root of the project?

jwage:
I think I am going to have to make a change to not use the real github API in the functional tests.

jwage:
I am working on fixing the github credentials error here: https://github.com/doctrine/doctrine-website/pull/263

SenseException:
> Could you please fix the coding styles

Coding styles are fixed now. :+1:

.scrutinizer.yml
```diff
@@ -0,0 +1,35 @@
+build:
+    nodes:
+        analysis:
+            environment:
+                php:
+                    version: 7.2
+                node:
+                    version: 6.14
```
SenseException: Do we need to use the latest LTS of 10 or doesn't it matter? Version 6 is currently maintenance only.

jwage:
I merged the change in https://github.com/doctrine/doctrine-website/pull/263 so you can rebase and the github credential failures should not happen for you anymore.

package.json
mmollick: I've updated this to use the 4 space indentation. In the future, this may change again if someone uses NPM rather than Yarn to manage this file. NPM will reset the file to two spaces but Yarn will respect the current indentation.

.scrutinizer.yml
mmollick: I've updated the version to the current LTS, 6.14 just happened to be the version I have installed locally. Any version that Webpack supports (currently >=6.11) should be sufficient.

.gitignore
```diff
@@ -10,3 +10,5 @@
 /.phpcs-cache
 /build-*
 /projects
+/node_modules
+.webpack-build
```
jwage: Add / to the beginning.

mmollick:
I'm getting a new error during the coverage test on Travis: `InvalidArgumentException: You must configure a Github http token.`.  It seems to run fine locally, any thoughts?

It also looks like scrutinizer doesn't like one of the changes from #263 with this new config.

mmollick:
This PR should be good to go. Scrutinizer is picking up on an issue introduced by #263 though.

jwage:
I will need to test this on the build server and make sure it has nodejs installed. What version should I install?

mmollick:
@jwage I would suggest either 8.x or 10.x, but anything over 6.11 will work
@SenseException That should be doable, we'll need to add some code splitting rules to webpack so we don't end up with masive bundle sizes and keep load times down.

mmollick:
All the frontend assets are now installed with Yarn/NPM and compiled by Webpack. Adding in the new JS and Sass to the build steps introduced a few changes to this PR:
- PurgeCSS has been added, this will comb through `.twig` and `.html` files and strip out unused CSS classes. With Bootstrap, Font-Awesome, and Algolia the CSS file sits at 255kb (unminified) but with PurgeCSS this brings the CSS down to 63.3kb (unminified). PurgeCSS will slow down the compilation times by about ~30 seconds, so it's skipped while running `npm run watch`. 
- A new `index.js` file was added to act as Webpack's entry point. This utilizes Webpack's built-in code splitting `import()` to create code bundles that will load asynchronously.
- Frontend assets (JS, CSS, and Webfonts) are now placed into a `/frontend` directory within the build directory. Previously the JS and CSS files would be overwritten by the source during the PHP build process. Alternatively, we can add the symlink/copying of the Webpack build after the PHP build process.
- Added a minimum Node version to the `package.json` to ensure a compatible version of NodeJS is used.

jwage:
Thanks. Sorry for the delay on my end. I plan on working on getting this merged soon.

mmollick:
@SenseException The latest changes are ready for review

mmollick:
>tooltip is not a function

This may be an issue with the order of the scripts loading. I'll have look into it further. Was this only on a specific page? And what browser did you notice this in? 

>Documentation

It sounds like the sidebar's JS isn't loading properly. I'll look into this as well.

>It seems that the build doesn't work when I install the packages with yarn with --prod. Are there dependencies from dev, that are still needed by prod?

You'll always want to do a full install of the Yarn dependencies without the `--prod` flag. The only time we'll be using any of the Node packages is during development and building so the difference between dev and prod dependencies comes down to semantics. `devDependencies` contain all the packages for build tools and `dependencies` contain the packages that eventually get compiled by webpack.

SenseException:
> This may be an issue with the order of the scripts loading. I'll have look into it further. Was this only on a specific page? And what browser did you notice this in?

No specific page. It appeared on the homepage, Community, Blog, Documentation... Mostly after a reload. The browser I used to test it was Chrome 70 and Firefox 63.

mmollick:
@SenseException Both issues have been resolved

source/js/index.js
```diff
@@ -0,0 +1,36 @@
+import(/* webpackChunkName: "sentry" */ '@sentry/browser').then(module => {
```
SenseException: Do we need Sentry for production?
We might need to get a dsn for ourselves too.

jwage:
@senseexception I just need to do a test build on the build server and make sure I have node installed then we should be g2g. I will have time this weekend to work on this.

source/js/index.js
mmollick: The DSN being used here is the same one currently in use on the Doctrine website. I assume it's owned by a contributor. If not we can remove Sentry for now

SenseException: Ping @jwage 
Is this still an active dsn? Do we currently use Sentry?

jwage:
I installed nodejs on the build server:

```
$ nodejs --version
v8.13.0
```

When I run the build process for this branch, it fails with this error `webpack: not found`

I think the deployer process should probably run `yarn install` after `composer install`. See here: https://github.com/doctrine/doctrine-website/blob/d0a1ee0b3be04c1ca3f51bf132fc6894e1ca22e5/lib/Deployer.php#L65

When I try to run `yarn install` I get this:

```
$ yarn install
The program 'yarn' is currently not installed. You can install it by typing:
apt install cmdtest
```

Sorry, I am a nodejs n00b. Can you advise on what other steps are needed to setup everything needed for the new build process?

mmollick:
Yarn will need to be installed alongside Node on the build server. Instructions for this can be found [here](https://yarnpkg.com/en/docs/install#debian-stable) (Instructions vary by OS). I'll make the change to the Deployer class to run `yarn install` after `composer install`.

jwage:
Thanks. I got yarn installed and it worked but the ./doctrine build-website command ran out of memory when copying all the files over for the static website. Are files being put in the build directory that don't need to be there by chance?

mmollick:
In the build directory it should just be creating a new `frontend` directory that contains the CSS, Javascript and webfonts output by Webpack. If the publish option isn't used this `frontend` directory will be a symlink to Webpacks output directory, but this symlinked directory still only contains the files Webpack outputs. In either case these directories should be well under 1MB. 

A more likely culprite would be Webpack and the plugins it's using. Webpack could be eating up a lot of memory especially with the PurgeCSS plugin. Can you try running `npm run build` on the server and see if it completes successfully?

jwage:
The npm commands run successfully, it is the PHP code that runs after that to build the static website. I need to look at why we have a memory leak there but something about the files needing to be built increased because it is processing many more files than it did before.

jwage:
I don't see anything that has drastically changed. This might have just been a pre-existing issue and the slight increase in files just barely pushed us over the limit. I need to investigate the underlying memory problem in https://github.com/doctrine/static-website-generator

jwage:
This is now live in production! I increased the memory limit for now while I work on improving the memory performance of some underlying libs.

Thanks so much for your hard work. This is a great foundation to keep building on top of.

SenseException:
@mmollick Thank you for your work to include package manager and webpack to our website. :+1:

jwage:
I am trying to write some new sass code and I put this inside of `source/styles/index.scss`:

```
.carbon-poweredby {
  color: black;
}
```

But it never ends up in the `/frontend/css/index.css` file that gets written. What am I doing wrong?

jwage:
I think it is because of the PurgecssPlugin. I am trying to style some HTML that is dynamically added to the DOM with JS from a 3rd party.

mmollick:
You can modify the PurgeCSS whitelist option on [this line](https://github.com/doctrine/doctrine-website/blob/8f503958dadd49ec87bf642dd091399e7d9944ce/webpack.config.js#L28). If theres classes from external sources we need to keep we can specify a regex expression that matches them. This is already being done for the classes Angolia's search uses.

Changing this value to `[/^ais/, /^carbon/]` should keep any classes that start with `carbon` as well.

jwage:
Thanks. I stumbled on that and got it working. I have much to learn :)

jwage:
@mmollick are you interested in helping me with this? https://github.com/doctrine/doctrine-website/pull/271

PR #267, opened 2018-11-05
Upgrade RST Parser.

jwage:
- Require PHP 7.2
- Update to Symfony 4.2
- Update `doctrine/rst-parser` to `0.1`
- Update `doctrine/skeleton-mapper` to `1.0`
- Update `doctrine/static-website-generator` to `1.0`
- `__toString()` was removed in `doctrine/rst-parser` so we have some changes in this PR to remove casting to string and instead we just call the `render()` method.
- `doctrine/rst-parser` had some bugs that caused invalid HTML markup to be produced and `doctrine/doctrine-website` had some CSS and JS that had to change after the upgrade.


SenseException:
This PR depends on a new release of doctrine/rst-parser

PR #270, opened 2018-11-27
Fix of Scrutinizer major issues

SenseException:
This PR is about fixing build breaking issues of Scrutinizer. Further improvements should be added here to make future PRs not getting blocked by changes from the past.


lib/Controllers/TeamController.php
```diff
@@ -35,6 +37,8 @@ public function contributor(string $github) : Response
     {
         $contributor = $this->contributorRepository->findOneByGithub($github);
 
+        assert($contributor instanceof Contributor);
```
Ocramius: Move the assertion into the repository method, and then you can make the return value there non-nullable

SenseException: Done.

lib/Repositories/ContributorRepository.php
```diff
@@ -6,12 +6,17 @@
...
+use function assert;
 
 class ContributorRepository extends BasicObjectRepository
 {
     public function findOneByGithub(string $github) : Contributor
     {
-        return $this->findOneBy(['github' => $github]);
+        $contributor = $this->findOneBy(['github' => $github]);
+
+        assert($contributor instanceof Contributor);
```
lcobucci: `assert()` is not executed in production environment (or shouldn't be). For dev purposes, this should be `assert($contributor instanceof Contributor || $contributor === null)` - which would properly represent the possibilities of the return of `findOneBy()`.

I do think that the `assert()` is not necessary, though. Mostly because we can have a `null` return from `findOneBy()` and, in order to prevent a `TypeException`, we should raise our own exception for this, like:

```php
public function findOneByGithub(string $github) : Contributor
{
    $contributor = $this->findOneBy(['github' => $github]);

    if (! $contributor instanceof Contributor) {
        throw new ???;
    }

    return $contributor;
}
```

SenseException: `assert()` is used through the whole project code. The PHP part is only executed when the static web-files are generated. What's the reason to handle it here differently?

I've added your suggestion for `assert()`.

lcobucci: @SenseException I don't know exactly what php config is being to generate the static files (@jwage would probably know). However, if it's using production config, the assertion won't even have its opcode created. Nevertheless the assertion you added is incorrect, since `null` is a valid return for `findOneBy()`.

lcobucci: If you do follow my suggestion regarding `null` scrutinizer will definitely mark this as a possible issue, though.

SenseException: `null` is a valid return type of `findOneBy()` but not of `findOneByGithub()`. My first commit made the return type of `findOneByGithub()` nullable to cover the one of `findOneBy()`. After the first review here, I made `findOneBy()` not nullable again and moved the `assert()` from the controller to the Repository.

Can you please take a look at the other usages of `assert()` in the project to see, if I am not properly using `assert()` here? Or is the usage of `assert()` in the project incorrect? Or what is the difference between this and the other usages of `assert()`? I'm trying to understand your point.

lcobucci: > `null` is a valid return type of `findOneBy()` but not of `findOneByGithub()`. My first commit made the return type of `findOneByGithub()` nullable to cover the one of `findOneBy()`. After the first review here, I made `findOneBy()` not nullable again and moved the `assert()` from the controller to the Repository.

@SenseException my point is that `assert()` is not called in "production mode" (a.k.a. `zend.assertions=-1` && `assert.exception=1`), thus we'll have a `TypeError` when `findOneBy()` returns `null` when this code is executed (which is possibly unexpected).

The assertion `assert($contributor instanceof Contributor || $contributor !== null);` is **incorrect** because `findOneBy()` has two possible returns: an instance of `Contributor` or `null`. When running that code in "development mode" (a.k.a. `zend.assertions=1` && `assert.exception=1`) we'll have an `AssertionException` for `null` values (which is possibly also unexpected).

> Can you please take a look at the other usages of `assert()` in the project to see, if I am not properly using `assert()` here? Or is the usage of `assert()` in the project incorrect? Or what is the difference between this and the other usages of `assert()`?

As we discussed back on doctrine/coding-standard#47, `assert()` should be used for development purposes (IDE, static analysis, etc), for "production" we have to have our won exceptions for handling unexpected things. Based on this, I'd categorise these as "invalid" usages for production settings (apart from this PR):

### File/path handling - could be solved by using SPL (which throws exceptions)

- https://github.com/doctrine/doctrine-website/blob/c8b8d511c7cc7881a8cded96b6b4aaacec3bf61a/lib/Assets/AssetIntegrityGenerator.php#L40
- https://github.com/doctrine/doctrine-website/blob/c8b8d511c7cc7881a8cded96b6b4aaacec3bf61a/lib/Assets/AssetIntegrityGenerator.php#L43
- https://github.com/doctrine/doctrine-website/blob/530f7f910e252d7fa8da15af88fdc9132e06975b/lib/Twig/MainExtension.php#L82
- https://github.com/doctrine/doctrine-website/blob/530f7f910e252d7fa8da15af88fdc9132e06975b/lib/Twig/MainExtension.php#L85
- https://github.com/doctrine/doctrine-website/blob/d0a1ee0b3be04c1ca3f51bf132fc6894e1ca22e5/lib/Deployer.php#L128
- https://github.com/doctrine/doctrine-website/blob/c8b8d511c7cc7881a8cded96b6b4aaacec3bf61a/lib/Projects/ProjectDataReader.php#L192

### Input validation - we should throw exceptions

- https://github.com/doctrine/doctrine-website/blob/4abe92371bde03fcf74d68ad202b86c7f1cee375/lib/RST/Directive/CodeBlockDirective.php#L95

The other assertions are correct IMHO :smile:

lcobucci: @SenseException I hope this clarifies my point and I'm quite sorry if I'm making a big deal out of this. I'm too passionate about `assert()` usage HAHAHAHA

SenseException: Thank you for your explanation. I was expecting `assert()` to catch `null` in development too, because of the not nullable return type `Contributor`. I'll add the expected exception and will do new commits for other cleanups.

SenseException:
@jwage You were too fast :smiley:

jwage:
Was something wrong with it? Seemed like a really simple change that was done.

SenseException:
See @lcobucci comment.

I also wanted to fix a few other major issues in this PR.

jwage:
I see. My mistake.

PR #271, opened 2018-11-30
Add carbon cpm ad to the website

jwage:
This PR adds the carbon cpm (1.00–$1.60 per 1000 page views) ad unit to the website. I will send in another PR with the CPC ad unit which pays 0.50–$1.20 per click.

### DIFF
--- a/templates/carbonad-standard.html.twig
+++ b/templates/carbonad-standard.html.twig
@@ -1,0 +1,1 @@
+<script async type="text/javascript" src="//cdn.carbonads.com/carbon.js?serve=CK7D553W&placement=wwwdoctrine-projectorg" id="_carbonads_js"></script>
--- a/source/404.html
+++ b/source/404.html
@@ -7,6 +7,12 @@ title: Four o four
 {% block content %}
     <h2>Page not found</h2>
 
+    <div class="row">
+        <div class="col">
+            {% include "carbonad-standard.html.twig" %}
+        </div>
+    </div>
+
     <p class="lead">The page you are looking for could not be found. Take a look at one of the pages below instead.</p>
 
     <ul>
--- a/source/index.html
+++ b/source/index.html
@@ -10,14 +10,26 @@
             <a href="{{ site.url }}/projects/dbal.html">Database Abstraction Layer (DBAL)</a> it is
             built upon.</p>
 
-            <p class="lead">
-                <a class="btn bg-doctrine-orange text-white btn-lg" href="{{ site.url }}/projects/doctrine-orm/en/current/tutorials/getting-started.html" role="button">Get Started</a>
-                <a class="btn bg-doctrine-dark-blue text-white btn-lg" href="{{ site.url }}/projects.html" role="button">View Projects</a>
-
-                <div class="btn btn-lg tidelift-badge">
-                    <a href="https://tidelift.com/subscription/pkg/packagist-doctrine-orm?utm_source=packagist-doctrine-orm&utm_medium=website" target="_blank" rel="noopener noreferrer"><img src="{{ get_asset_url('/images/tidelift-icon.png', site.url) }}" class="tidelift-icon" /> <span>Get professional support for Doctrine</span></a>
+            <div class="row">
+                <div class="col-12">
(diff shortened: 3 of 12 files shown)


source/js/carbonads.js
```diff
@@ -0,0 +1,174 @@
+export default function() {
+  var _native = (function () {
```
jwage: @mmollick This javascript was provided by carbonads and they call _native from the global scope here https://github.com/doctrine/doctrine-website/pull/271/files#diff-b94a2300afd9b8a8da845e9e7bc1a47cR201

What is the proper way to do this to make something available to be called?

source/js/index.js
```diff
@@ -14,6 +14,10 @@ import(/* webpackChunkName: "search" */ './search').then(module => {
     module.default(projectSlug, versionSlug, searchBoxSettings);
 });
 
+import(/* webpackChunkName: "carbonads" */ './carbonads').then(module => {
+    var _native = module.default();
```
jwage: I think this is wrong.

source/js/carbonads.js
```diff
@@ -0,0 +1,174 @@
...
+
+      var jsonUrl = 'https://srv.buysellads.com/ads/' + zone + '.json?callback=_native_go'
+      if (_options['placement'] !== '') {
+        jsonUrl += '&segment=placement:' + _options['placement']
+      }
+      if (_options['ignore'] === 'true') {
+        jsonUrl += '&ignore=yes'
+      }
+
+      var srv = document.createElement('script')
```
Ocramius: This is literally injecting third party JS (even if JSON) into our pages.

This requires adding a disclaimer to all our pages due to tracking (EU norms).

Do we really want to go down this path?

source/js/carbonads.js
```diff
@@ -0,0 +1,174 @@
...
+      srv.src = jsonUrl
+      document.getElementsByTagName('head')[0].appendChild(srv)
+    }
+
+    var carbon = function (e) {
+      var srv = document.createElement('script')
+      srv.src = '//cdn.carbonads.com/carbon.js?serve=' + e['carbonZoneKey'] + '&placement=' + e['placement']
```
Ocramius: More questions for the carbon folks: why two domains?

jwage:
@Ocramius I took a small poll on Twitter to see what the sentiment of our users might be. Small sample size but it doesn't seem to be a big deal. https://twitter.com/jwage/status/1063440052112494592

We chatted about it in Gitter and there was pushback similar to yours. Questions were asked about GDPR and what data they collect. We determined, after talking to CarbonAds, that they don't track user data so it does not require GDPR consent.

The ability to style the ads to make it unobtrusive for the user plus the potential $ that this might generate to support the team seems to make it worth it.

source/js/carbonads.js
jwage: We may not want to, some conversations in gitter led to wanting to see what it would look like so I am currently in hack something together to see how it might look mode.

Ocramius:
Ok, let's give it a try and see the reactions then, but at the first smell of security concerns this is to be removed.

source/js/carbonads.js
jwage: I think carbonads was bought by buysellads.com

templates/carbonad-standard.html.twig
```diff
@@ -0,0 +1,3 @@
+<div class="carbonad-standard">
+    <script async type="text/javascript" src="//cdn.carbonads.com/carbon.js?serve=CK7D553W&placement=wwwdoctrine-projectorg" id="_carbonads_js"></script>
```
SenseException: Is it possible to use `integrity` for carbon ads scripts?

jwage: I will find out how often they change the javascript.

jwage: We could use it, but every time they change the javascript, it would break.

SenseException: I don't find anything about Carbonads script versions in a quick Google search.

jwage: I talked to carbon folks and they said the js doesn't change that often and that we can use the integrity attribute so I will add it.

jwage: Ok more info. The script is updated every 1 to 2 months and they don't recommend using sri hashes.

jwage:
@Ocramius @SenseException This is ready to deploy now so we can see the impact it has.

PR #274, opened 2018-12-18
Default project version branchName and slug to name when it is not present

jwage:
I am going through the projects and updating the .doctrine-project.json and wanted to set these defaults in code so I can reduce the amount of data that has to be manually configured in .doctrine-project.json

### DIFF
--- a/lib/Model/ProjectVersion.php
+++ b/lib/Model/ProjectVersion.php
@@ -53,8 +53,8 @@ class ProjectVersion
     public function __construct(array $version)
     {
         $this->name       = (string) ($version['name'] ?? '');
-        $this->branchName = (string) ($version['branchName'] ?? '');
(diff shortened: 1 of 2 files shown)


jwage:
The travis failure is a github/travis problem not related to the code changes. Everything is passing locally. I am going try 1 more restart and then merge.

jwage:
Actually, this might be a new problem. Investigating.

jwage:
This was actually a problem I introduced in https://github.com/doctrine/doctrine-website/pull/251 but we did not see it due to cache in travis. I made a change in that PR that removed support that gave the ability for a project and the docs to live in separate repositories. I fixed it by moving the doctrine1 docs to the same repo https://github.com/doctrine/doctrine1/pull/83. That was the only odd repo that did it that way so the Doctrine website build process doesn't have to support that oddity anymore.

Issue #166, opened 2018-05-04
Dead links found on website


SenseException:
By using wget as a spider, I checked the website for 404 / errors and retrieved the following list:

[wget.log.gz](https://github.com/doctrine/doctrine-website/files/1973796/wget.log.gz)

At the end of the list are 305 broken links listed. Most of them are from templates using a placeholder {{url}}, which may not be usable by wget, but this needs to be confirmed first.

The used wget command to create the log-file:

`wget --spider -r -nd -nv -l 10 -w 2 -o wget.log  https://www.doctrine-project.org`

I used 10 levels `-l` for my check, which made wget run ~8 hours and a waiting time `-w` of two seconds between the downloads for the web server. A higher level may show more broken links.

Related issues:

- #164

SenseException:
Some external links used in the documentation aren't 404, but lead to deprecated or outdated content. These links aren't listed in the url-list.

deltab:
What about broken links found elsewhere? e.g. https://doctrine-orm.readthedocs.io/en/latest/reference/annotations-reference.html from search engine results.

SenseException:
https://doctrine-orm.readthedocs.io isn't in use anymore. Please use https://www.doctrine-project.org for all the Doctrine projects.

vladyslavstartsev:
I've found that there are a lot of broken links due to jira

something like this
http://www.doctrine-project.org/jira/browse/DCOM-129

what I can try to do is to find a link like the above one and change it to https://github.com/doctrine/common/issues/428

is that a good idea? should I try?

jwage:
Yes, we would accept PRs to docs to update links :+1:

SenseException:
@vladyslavstartsev The PR doctrine/orm#7385 covers fixes for links of doctrine/orm, but I haven't checked the other projects yet.

Ocramius:
Is there a grep syntax that worked for you here? Error messages seem to be on a new line each...

SenseException:
No, it was just the command above. After some links were fixed since the creation of the log, we can consider another scan with a different tool.

vladyslavstartsev:
I used level 20
`wget --spider -r -nd -nv -l 20 -w 2 -o wget.log https://www.doctrine-project.org`
[wgetlog.tar.gz](https://github.com/doctrine/doctrine-website/files/2872915/wgetlog.tar.gz)

SenseException:
[...]

PR #277, opened 2019-01-06
remove dead links

vladyslavstartsev:
Mostly it was 
- jira to github issues
- symfony-project to symfony

as discussed here https://github.com/doctrine/doctrine-website/issues/166#issuecomment-451611200

also I don't know what to do with 
- "Please report any issues you may have with the update on the mailing
list or on [Jira](http://www.doctrine-project.org/jira/browse/DDC)."
- www.doctrine-project.org/change_log (for example www.doctrine-project.org/change_log/1_2_0_BETA3 ) which is dead but there is no corresponding live version
- http://www.doctrine-project.org/download same as previous one
- https://www.doctrine-project.org/2008/10/02/doctrine-1-1-development-begins.html is empty blog , what should be done here?

---
note to self:
- update all doctrine-project.org non-https to https in other PR
- update links from https://github.com/doctrine/doctrine2 to https://github.com/doctrine/orm in other PR


jwage:
Thanks! This is awesome. The other ones I don't think we have an alternative for right now.

PR #278, opened 2019-01-08
http to https

vladyslavstartsev:
Websites that where moved

- github.com
- php.net to secure.php.net
- www.doctrine-project.org
- orientdb.org
- packagist.org
- twitter.com
- martinfowler.com
- sensiolabs.com
- en.wikipedia.org
- www.mongodb.org to www.mongodb.com
- amazon.com
- 3v4l.org
- php.budgegeria.de
- framework.zend.com
- www.sitemaps.org
- schema.org
- www.w3.org
- www.atlassian.com
- magentocommerce.com to magento.com
- mysql.com
- slideshare.net
- groups.google.com

moved non-www doctrine-project.org to version with www

### DIFF
--- a/source/blog/2008-06-13-php-net-style-api-documentation-lookups.md
+++ b/source/blog/2008-06-13-php-net-style-api-documentation-lookups.md
@@ -8,7 +8,7 @@ permalink: /2008/06/13/php-net-style-api-documentation-lookups.html
 <p>
 
 I have been using php.net style API lookups ever since I can remember.
-It is very nice being able to simply go to http://php.net/array\_walk
+It is very nice being able to simply go to https://secure.php.net/array\_walk
 and have it pull up that function in the PHP API Documentation. So,
 tonight i have implemented this functionality in to the Doctrine
 website. Below are some examples of how you can quickly access the API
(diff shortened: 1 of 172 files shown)


source/blog/2008-06-13-php-net-style-api-documentation-lookups.md
```diff
@@ -8,7 +8,7 @@ permalink: /2008/06/13/php-net-style-api-documentation-lookups.html
 <p>
 
 I have been using php.net style API lookups ever since I can remember.
-It is very nice being able to simply go to http://php.net/array\_walk
+It is very nice being able to simply go to https://secure.php.net/array\_walk
```
morozov: I think the original idea was that you can go to the definition of `array_walk()` just by typing "php.net/array_walk" which is indeed simple. However, typing "https://secure.php.net/array_walk" is not that simple (people usually don't do that). Would it make sense to reword the original statement and get rid of HTTP/HTTPS there entirely since it's irrelevant?

jwage: This blog post is old and irrelevant now. We don't have the equivalent functionality on the website now.

PR #287, opened 2019-02-08
Update highlight.php dependency

allejo:
Hai there :wave:

I currently maintain highlight.php and noticed this project was locked to v9.12.0.4. You may run into https://github.com/scrivo/highlight.php/issues/33 if you use PHP 7.3 because of this Composer constraint.

A PR to update the composer files is probably weird but I'm just passing along trying to prevent someone from getting a headache because of this bug :sweat_smile:


jwage:
Thanks!

PR #296, opened 2019-05-03
Update yarn dependencies

SenseException:
I've updated some of the yarn dependencies for the website.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,14 +14,14 @@
         "test": "tests"
     },
     "dependencies": {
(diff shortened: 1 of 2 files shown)


Ocramius:
Does it make sense to install @dependabot here?

SenseException:
I have no experience with Dependabot. We should check out its possibilities and take a look if it can help us keep the website up to date without breaking major things.

Ocramius:
I use it on https://github.com/Roave/SecurityAdvisoriesBuilder - we'd need to depend on the CI process to verify things though.

PR #332, opened 2019-12-20
Compress logo and remove duplicate

pichfl:
The current logo used on the Doctrine website was not optimized for file size and was duplicated in two places in the source files.

This PR compresses the logo to 1/6 of the original file size while keeping it optically identical and re-uses the logo file in the main layout.


SenseException (approved):

SenseException:
Thank you @pichfl

PR #333, opened 2020-01-15
Remove memory-limit in build-website-data command

SenseException:
The build/deployment for the website is exiting with the error message

```
PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried   
  to allocate 307200 bytes) in vendor/doctrine/static-website-generator/lib/SourceFile/SourceFile.php on line 131
```

This occurs because of a limitation of memory added with `ini_set` in the past. Without the limit, the build proceeds.


SenseException:
I have to remove the `use` of the function.

SenseException:
No, but maybe it had a limit once on cli.

Issue #324, opened 2019-09-27
Link to Slack invite is outdated


rbairwell:
On https://www.doctrine-project.org/community/  under the heading of "Slack" it links to https://www.doctrine-project.org/slack which then redirects to https://doctrine.slack.com/join/shared_invite/enQtNjE5NDU1MTIyMzU5LTY2Mjk5NDlhODEzZDRhYmY5ZjNjODRhNTdiOGI1ZGMzNDQwOTk0YzA1OGNkOWRlZjIyZjdkYTNkYmQ4N2RmMTY - however  Slack reports "This invite link is no longer active.
Please check with the person who shared it with you to see if there's a new link available."

Could the redirect be updated or the link to Slack removed?

SenseException:
Seems like https://www.doctrine-project.org/slack was already fixed or it was just a small inconvenience. Thank you for reporting it. :+1:

VladyslavChernyshov:
> Seems like https://www.doctrine-project.org/slack was already fixed or it was just a small inconvenience. Thank you for reporting it. 👍

It is outdated again.

PR #345, opened 2020-04-04
Doctrine migrations 3.0 post 

goetas:
Not sure it makes sense this pull request, but this re-publishes on the official doctrine blog the [article](https://www.goetas.com/blog/released-doctrinemigrations-30-alpha/) I wrote for the doctrine/migrations 3.0-alpha release.

I have mentioned the original post at the bottom of the article and added the canonical url link to avoid google penalties for duplicate content.

If you think is not a good idea, feel free to close the pull request.


malarzm:
I think it's the right place for such a blog post :) Quickly looking at the files I'd say that there's no point in adding doctrine logo, there should be something already in the repository :)

goetas:
>Quickly looking at the files I'd say that there's no point in adding doctrine logo, there should be something already in the repository :) 

Logo removed! was just a copy/paste mistake :)

goetas:
I'm not familiar with this project, do I have to edit the `BlogPostHydrator` and `BlogPost` classes adding the "canonical" property?

templates/layouts/layout.html.twig
```diff
@@ -26,6 +26,10 @@
             <meta name="robots" content="noindex">
         {% endif %}
 
+        {% if page.canonical %}
```
SenseException, line 29:
```suggestion
        {% if page.canonical is defined and page.canonical %}
```

I did some tests and this change was enough to make the canonical appear when is was provided in the blog post.

goetas: fixed

SenseException:
The build is failing because of website configurations in other branches. Probably just #324. The blog post will appear only with a working build. After the fixes there are done and the build works, I'll test, approve + merge the blog post.

goetas:
Thanks

SenseException:
Thank you

PR #347, opened 2020-06-04
Prepare build for PHP 7.4

SenseException:
A PHP update should be done for the website. This PR is a first step to prepare the build for the usage of PHP 7.4.

* Update of dependencies
* Fix of coding styles
* Support for 7.2 and 7.4 during the transition


lib/Cache/CacheClearer.php
```diff
@@ -68,6 +68,6 @@ public function clear(string $buildDir) : array
      */
     protected function glob(string $pattern) : array
     {
-        return glob($pattern);
+        return array_filter((array) glob($pattern));
```
SenseException, line 71: A phpstan error tells that `glob()` seems to be able to return `false` and also `array<int, false>`. This is why I created this strange filter-typecast construct.

phpstan.neon.dist
```diff
@@ -7,3 +7,7 @@ parameters:
     paths:
       - lib
       - tests
+    ignoreErrors:
+      # TODO Use newer Doctrine packages which are using T and TKey
```
SenseException, line 11: Currently doctrine-website is using outdated collection dependencies where `T` and `TKey` aren't available yet. The outdated version is because doctrine/skeleton-mapper is also using outdated dependencies.

SenseException:
This is all I can do on the doctrine-website side so make it work with PHP 7.4.

I need to do further updates for

* doctrine/skeleton-mapper
* doctrine/rst-parser

Because of doctrine/rst-parser there are tests failing.  doctrine/skeleton-mapper seems to be the culprit that keeps dependencies on an outdated version. For now I used `ignoreErrors` in PHPStan for the cases where I currently can't define types.

I would prefer to handle the implementation of the updated skeleton-mapper and rst-parser in separate PRs and make the PHP 7.4 builds then mandatory.

.travis.yml
```diff
@@ -39,22 +37,65 @@ jobs:
...
+      env: COVERAGE
+      before_script:
+        - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{.disabled,}
+        - if [[ ! $(php -m | grep -si xdebug) ]]; then echo "xdebug required for coverage"; exit 1; fi
+      script:
+        - ./bin/console --env=test build-all
+        - ./vendor/bin/phpunit --coverage-clover clover.xml
+      after_script:
+        - wget https://scrutinizer-ci.com/ocular.phar
+        - php ocular.phar code-coverage:upload --format=php-clover clover.xml
```
greg0ire, line 86: Please use codecov here, you will have to rebase to get an example for 7.2

.travis.yml
```diff
@@ -39,22 +37,65 @@ jobs:
...
+      script: ./vendor/bin/phpcs
+
+    - stage: Code Quality
+      php: 7.4
+      env: STATIC_ANALYSIS
+      install: travis_retry composer install --prefer-dist
+      script: ./vendor/bin/phpstan analyse
+
+  allow_failures:
+    - php: 7.4
```
greg0ire, line 101: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

lib/Model/Event.php
```diff
@@ -40,16 +40,16 @@ final class Event implements LoadMetadataInterface
     /** @var EventLocation|null */
     private $location;
 
-    /** @var EventSponsors */
+    /** @var EventSponsors<EventSponsor> */
```
greg0ire, line 43: This looks a bit weird… I mean you can't have `EventSponsors<SomeRandomClass>`, can you?

.travis.yml
SenseException: Yes, I didn't make a rebase yet. Done.

lib/Model/Event.php
SenseException: phpstan was complaining about that elements have to be `EventSponsor`. It doesn't do it anymore so I guess one of my fixes that came after might cover those ones.

SenseException:
@greg0ire Builds are green again.

lib/Model/EventSchedule.php
```diff
@@ -10,16 +10,20 @@
...
+/**
+ * @template-extends AbstractLazyCollection<int, \Doctrine\Website\Model\EventScheduleSlot>
+ */
 final class EventSchedule extends AbstractLazyCollection
 {
     /** @var mixed[] */
     private $event;
 
-    /** @var EventSpeakers */
+    /** @var EventSpeakers<EventSpeaker> */
```
greg0ire, line 21: Does this really have to be that way? Phpstan should be able to figure that `EventSpeakers` elements can't be anything but `EventSpeaker` instances, shouldn't it?

SenseException: @greg0ire Done. It became obsolete because of `@template-extends` on the class.

SenseException:
@greg0ire Thank you for your time.

PR #348, opened 2020-06-20
Use CodeCov instead of Scrutinizer

greg0ire:

PR #349, opened 2020-06-23
Fix of CodeCov link in README

SenseException:
@greg0ire I've missed a wrong link in your PR #348.


greg0ire:
Thanks for taking care of that!

PR #351, opened 2020-07-23
Regular builds to make broken .doctrine-project.json visible

SenseException:
A misconfigured .doctrine-project.json file in Doctrine projects tend to break builds which prevents all project documentations to be released in case of new changes. As a current first step this action was created to regularly run the `build-docs` command to make broken website configs in the different projects visible.


SenseException:
Thank you for the review @ostrolucky

PR #362, opened 2020-09-25
Lazy load commands

greg0ire:
Making the command name available without instantiating it makes it
possible to find the right command without instantiating all of them.

### DIFF
--- a/lib/Commands/ClearBuildCacheCommand.php
+++ b/lib/Commands/ClearBuildCacheCommand.php
@@ -15,6 +15,9 @@ use function sprintf;
 
 class ClearBuildCacheCommand extends Command
 {
+    /** @var string */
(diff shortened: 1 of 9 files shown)


ostrolucky (approved):

lib/Commands/ClearBuildCacheCommand.php
```diff
@@ -15,6 +15,9 @@
 
 class ClearBuildCacheCommand extends Command
 {
+    /** @var string */
+    protected static $defaultName = 'clear-build-cache';
```
SenseException, line 19: `setName('clear-build-cache')` is still there.

SenseException (changes requested):

SenseException (approved):

PR #365, opened 2020-09-29
Install no dev dependencies for website-configs build

SenseException:
`Symfony\Component\ErrorHandler\Debug` is currently breaking the build because of notices. Because this build is supposed to check only website configs regularly, the dev dependencies of composer and therefore `Debug` doesn't need to be installed.


greg0ire (approved):

SenseException:
Thanks again for pointing to the `--no-dev`.

PR #367, opened 2020-10-03
Update Doctrine Coding Standard to 8.1

### LINKED ISSUE #366: Cannot install with Composer 2
```
$ composer --version
Composer version 2.0.0-RC1 2020-09-10 15:39:45

$ composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - dealerdirect/phpcodesniffer-composer-installer is locked to version v0.5.0 and an update of this package was not requested.
    - dealerdirect/phpcodesniffer-composer-installer v0.5.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 2
    - dealerdirect/phpcodesniffer-composer-installer v0.5.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - doctrine/coding-standard 5.0.1 requires dealerdirect/phpcodesniffer-composer-installer ^0.5.0 -> satisfiable by dealerdirect/phpcodesniffer-composer-installer[v0.5.0].
    - doctrine/coding-standard is locked to version 5.0.1 and an update of this package was not requested.

Use the option --with-all-dependencies to allow upgrades, downgrades and removals for packages currently locked to specific versions.

You are using a snapshot build of Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report an issue to them to ask them to support Composer 2. To work around this you can run Composer with --ignore-platform-req=composer-plugin-api, but this may result in broken plugins and bigger problems down the line.
```

morozov:
Additionally, fix invalid mock setup which otherwise would cause the following:
```
There was 1 warning:

1) Doctrine\Website\Tests\Projects\ProjectVersionsReaderTest::testReadProjectVersions
Method getRepositoryTags may not return value of type string, its return declaration is ": array"
```

Fixes #366.

As for the container failure, I ended up removing the service options: regardless of the timeout, the container failed as unhealthy after the first attempt with the timeout of 2s. The defaults seem saner.


morozov:
@SenseException this is now ready for review.

SenseException (approved):

PR #370, opened 2020-10-11
Adapt contribution workflow to the default branch approach

SenseException:
This PR adapts the contributor workflow to the information about using the default branch and the up-merge.

It still lacks some description about when to use the default branch (bugfixes, features only when there's no tagged release yet) or when to use the next version branch above. I'll need some feedback about how to make this piece of explanation more understandable.


source/contribute.rst
```diff
@@ -57,31 +57,42 @@ Initial Setup
...
--  Adjust your branch to track the doctrine master remote branch, by
-   default it'll track your origin remote's master:
+Branching from the default branch
+---------------------------------
 
-.. code-block:: console
+New pull requests are created with the repository's default branch as base branch.
+The default branch is the branch you see when you enter the repository page on GitHub.
+
+.. image:: ../images/default-branch.png
```
SenseException, line 66: The image is visible after the website is build.

source/contribute.rst
```diff
@@ -57,31 +57,42 @@ Initial Setup
...
+   :alt: The default branch
+   :style: margin-bottom: 20px
 
-    $ git config branch.master.remote doctrine
+In this ORM example, it's the branch with the name **2.7**. The branch name reflects the
```
greg0ire, line 70: Maybe the ORM is not the best example… naming a branch without a `.x` is kinda deprecated since the automatic releases workflow creates them with that suffix. Consider using the DBAL as an example.

source/contribute.rst
```diff
@@ -57,31 +57,42 @@ Initial Setup
...
 
-Keeping your master up-to-date!
--------------------------------
+Newly introduced changes to 2.7 will be up-merged at a later point in time
+to newer version branches (e.g. 2.8, 3.0). This way you don't have to
+re-introduce a new fix or feature of 2.7 with another pull request to
+the other version branches.
 
-Once all this is done, you'll be able to keep your local master up to
+Keeping the default branch up-to-date!
```
greg0ire, line 78: Maybe we should remove that paragraph entirely, or reduce it to `git fetch`. IMO, people should use `git switch --create my-branch origin/2.8.x`, not `git switch --create my-branch 2.8.x`

source/contribute.rst
```diff
@@ -107,43 +118,40 @@ Primary Branches
...
+**doctrine/2.7** is the branch where the source code of **HEAD**
 always reflects the latest version. Each released stable version will be
 a tagged commit in a **doctrine/\*** branch. Each released unstable
-version will be a tagged commit in the **doctrine/master** branch.
+version will be a tagged commit in the **doctrine/2.7** branch.
 
-    **NOTE** You should never commit to your forked **origin/master**.
-    Changes to **origin/master** will never be merged into
-    **doctrine/master**. All work must be done in a **topic branch**,
+    **NOTE** You should never commit to your forked default branch (**origin/2.7**).
```
greg0ire, line 131: Maybe we should advise people to just remove their local copy of 2.7, that way they will not be tempted to do so.

source/contribute.rst
SenseException: Good point. I've changed it to DBAL.

source/contribute.rst
SenseException: People using the default branch name of the fork is because of convenience. They probably didn't read this docs.

greg0ire (approved):

source/contribute.rst
SenseException: I've reduced it to `git fetch`.

Issue #369, opened 2020-10-07
Add support for defining version tags instead of branches


morozov:
There's no much point in keeping the branches that correspond to the no longer supported versions in the repository. Contributors can create pull requests targeted at the branches that are no longer supported. The only reason we currently keep the branches in the DBAL is for the website.

This won't be necessary if the website can identify a no longer supported branch by its latest tag.

SenseException:
This was mentioned internally in Slack and using tags instead of branches is going to be a new additional configuration in `.doctrine-project.json` files.

I considered to fix doctrine/dbal#3338 in the older branches first, because a waiting release of doctrine/rst-parser would break the build there. This would still need branches to be available in the repository, but I scrapped this idea. My current plan is to skip errors, if possible, for older unsupported versions.

After the default branch and build changes for the website are done, I can pick this one.

morozov:
So far, I see that the `ProjectVersion` model already supports tags as such:
https://github.com/doctrine/doctrine-website/blob/2bec905fe15d5bed9231082749d0600ac52c519b/lib/Model/ProjectVersion.php#L69-L71
and the notion of the latest tag:
https://github.com/doctrine/doctrine-website/blob/2bec905fe15d5bed9231082749d0600ac52c519b/lib/Model/ProjectVersion.php#L168
Although I don't see any tests that would populate the tags from `.doctrine-project.json` nor I was able to find the projects that use tags in their project file. On the other hand, given that the website builder checks out the Git repo of the project anyways it could populate the model with the tags retrieved from Git.

SenseException:
Tags are currently grouped with a version from the config file of a project (slug). If `branchName` becomes optional and the checkout process supports tags too, we could introduce tags for the documentation and don't need to have a branch for every old project version.

PR #372, opened 2020-11-01
Add support for defining version tags instead of branches

SenseException:
Handles #369

### DIFF
--- a/lib/Projects/ProjectGitSyncer.php
+++ b/lib/Projects/ProjectGitSyncer.php
@@ -8,6 +8,7 @@ use Doctrine\Website\Github\GithubClientProvider;
 use Doctrine\Website\ProcessFactory;
 use Github\Api\Repo;
 
+use function escapeshellarg;
 use function is_dir;
 use function sprintf;
 
@@ -45,8 +46,8 @@ class ProjectGitSyncer
 
         $command = sprintf(
             'git clone https://github.com/doctrine/%s.git %s',
-            $repositoryName,
-            $repositoryPath
+            escapeshellarg($repositoryName),
+            escapeshellarg($repositoryPath)
         );
 
         $this->processFactory->run($command);
@@ -63,8 +64,8 @@ class ProjectGitSyncer
     {
         $command = sprintf(
             'cd %s && git clean -xdf && git checkout origin/%s',
-            $this->getRepositoryPath($repositoryName),
-            $branchName
+            escapeshellarg($this->getRepositoryPath($repositoryName)),
+            escapeshellarg($branchName)
         );
 
         $this->processFactory->run($command);
@@ -74,8 +75,8 @@ class ProjectGitSyncer
     {
         $command = sprintf(
             'cd %s && git clean -xdf && git checkout tags/%s',
-            $this->getRepositoryPath($repositoryName),
-            $tagName
+            escapeshellarg($this->getRepositoryPath($repositoryName)),
+            escapeshellarg($tagName)
         );
 
         $this->processFactory->run($command);
@@ -85,7 +86,7 @@ class ProjectGitSyncer
     {
         $command = sprintf(
             'cd %s && git clean -xdf && git fetch origin',
-            $this->getRepositoryPath($repositoryName)
+            escapeshellarg($this->getRepositoryPath($repositoryName))
         );
 
         $this->processFactory->run($command);
--- a/lib/Projects/ProjectGitSyncer.php
+++ b/lib/Projects/ProjectGitSyncer.php
@@ -70,6 +70,17 @@ class ProjectGitSyncer
         $this->processFactory->run($command);
     }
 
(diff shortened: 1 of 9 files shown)


SenseException:
I started to test a PoC with a temporary repository yesterday. If this is working, I'll finalize this draft and make a PR out of it.

SenseException:
If no `branchName` is given or set to `null`, the latest tag will be used for the build of the docs.

lib/Projects/ProjectGitSyncer.php
```diff
@@ -70,6 +70,17 @@ public function checkoutBranch(string $repositoryName, string $branchName): void
         $this->processFactory->run($command);
     }
 
+    public function checkoutTag(string $repositoryName, string $tagName): void
+    {
+        $command = sprintf(
+            'cd %s && git clean -xdf && git checkout tags/%s',
+            $this->getRepositoryPath($repositoryName),
+            $tagName
```
morozov, line 78: Although it will be used only for doctrine-owned projects, it still makes sense to implement proper escaping of the command line arguments.

morozov (commented):
Thank you, @SenseException! Looking forward to being able to use this.

lib/Projects/ProjectGitSyncer.php
SenseException: Good point. Done.

greg0ire (commented):
> This was mentioned internally in Slack and using tags instead of branches is going to be a new additional configuration in .doctrine-project.json files.

So how does it work? Should we document that somewhere?

SenseException:
Omit `branchName` or set it to `null`. Then the newest tagname will be used ("newest" as the used method name suggested and manual tests showed).

> Should we document that somewhere?

The documentation has no documentation :thinking: 
I should start to document `.doctrine-project.json`. At least as much as I know about it. I would like to create a separate PR for that.

lib/Projects/ProjectGitSyncer.php
```diff
@@ -45,8 +46,8 @@ public function initRepository(string $repositoryName): void
 
         $command = sprintf(
             'git clone https://github.com/doctrine/%s.git %s',
-            $repositoryName,
-            $repositoryPath
+            escapeshellarg($repositoryName),
+            escapeshellarg($repositoryPath)
```
greg0ire, line 50: I see the processFactory below uses `symfony/process`, which has support for
[escaping arguments](https://github.com/symfony/process/blob/240e74140d4d956265048f3025c0aecbbc302d54/Process.php#L177-L178):

```
     *   $process = Process::fromShellCommandline('my_command "$MY_VAR"');
     *   $process->run(null, ['MY_VAR' => $theValue]);
```

That would imply adding an extra argument for processFactory som that this can
be changed to

```php
$this->processFactory->run(
    'git clone "https://github.com/doctrine/$REPOSITORY_NAME.git" $REPOSITORY_PATH',
    null,
    ['REPOSITORY_NAME' => $repositoryName, 'REPOSITORY_PATH' => $repositoryPath]
);`
```

morozov: It looks like unnecessary coupling to the Process API to mo.

greg0ire: I think it's fair to say anything that deals with processes will have an associative array at some point, just like any library that deals with prepared SQL statements will have such arrays for binding named parameters. And the `ProcessFactory` abstraction is here to help making sure we can still ditch `symfony/process` and use something else while keeping that API, isn't it?

morozov: In this code, you just need to build a shell command which later will be executed. Building an abstraction over another abstraction looks like overkill and premature generalization.

greg0ire: Is it that premature though? There are already many calls to `run` combined with `sprintf` in the codebase, and that PR adds more of them. Looks more overdue than premature to me.

https://github.com/doctrine/doctrine-website/search?q=this-%3EprocessFactory-%3Erun

You will note that existing calls neglect to do the escaping, and rely on a `$command` OTV most of the times. Moreover, we already have an abstraction (`ProcessFactory` built over `symfony/process`).

morozov: @greg0ire do you want to discuss this as part of this PR?

greg0ire: I don't require we do that change as part of this PR, but it's still a good occasion to discuss it. Feel free to merge as is, I don't want this to be a blocker, and I might do the refactoring myself in a subsequent, so that the benefits are easier to spot.

greg0ire (approved):
Looks OK as is.

lib/Projects/ProjectGitSyncer.php
SenseException: It's easy to go off track with a website PR. There have to be a few more changes, but one small step at the time.

Thanks for the reviews.

PR #376, opened 2020-11-19
Missing headline in some blog posts

SenseException:
While headlines are shown in the blog post list, in some cases they aren't shown in the blog post itself. @stof found out that this is due to the current `slug` usage in the route. This PR is about covering this missing headline bug.

### DIFF
--- a/config/routes.yml
+++ b/config/routes.yml
@@ -210,3 +210,5 @@ parameters:
             defaults:
                 menuSlug: blog
                 layout: blog-post
+            requirements:
+                slug: .+


greg0ire (approved):
So since the dot were not allowed previously, the route did not match… why does the blog post display anyway then? Instead of producing a 404?

stof:
@greg0ire these routes are not about the website itself (which is a static website) but about the static generation, to add additional metadata. As it did not match, only the basic metadata (the one in the front matter) was used.

greg0ire:
Ok, it makes a lot more sense now.

PR #385, opened 2021-02-05
Add syntax highlighting in the DBAL 3.0 blog post

stof:


stof:
Looks like the actions workflow does not support PRs coming from forks, which don't have access to secrets.

stof:
If a read-only access is enough for that build, using `${{ github.token }}` in the workflow should work.

SenseException:
> Looks like the actions workflow does not support PRs coming from forks, which don't have access to secrets.

After #386 gets merged, you can rebase your branch. Thanks again for pointing out the github.token.

The blog pages don't seem to support the GitHub code highlighting notation. There are some blog posts using `~~~~ {.sourceCode .php}`, but this doesn't seem to have any effect either. I need to take a look how the code highlighting is handled in markdown by the website.

SenseException (approved):
I've checked how the website handles your changes. The third party library already in use creates a `pre`-`code` combination where the `php` in your markdown generates a CSS class in the code tag like `<code class="language-php">`. This can be used in combination with [prismjs](https://prismjs.com). Prism is not in use on the website yet, but can be added in another PR.

I'll merge your changes and will introduce prismjs in a new PR. After prismjs got introduced, your changes will result in syntax highlighting.

PR #389, opened 2021-02-24
Add paragraph about squashing

greg0ire:
We keep repeating the same things to contributors, having a resource
specifically about this would be nice.

### DIFF
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -195,6 +195,73 @@ this in case you want to read more about this:
 - `Commit message style for git <https://commit.style/>`_
 - `A note about git commit messages <https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>`_
 
+To squash or not to squash
+--------------------------
+
+The best way to avoid having to squash anything in the first place is to
+amend your last commit if that's indeed where your extra change is meant
+to go. That being said, sometimes you end up with many commits and it's
+too late for that. Some other times, code review has already started and
+it can be better not to touch already reviewed commits. You can signal
+that they should ultimately be squashed by using ``git commit
+--fixup=HEAD``, which will also spare you the creation of a commit
+message since it will reuse the previous one.
+
+Now let's say that code review is finished, or that it hasn't started,
+and that you want to squash some commits.
+
+If you are in the fairly simple case where you want squash all your
+commits into one, you can take `the following steps described in the
+manual
+<https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing>`_
+to achieve that.
+
+If you are in a more complex case where you would very much like to keep
+your commits separate, there are other solutions.
+To take a specific example, let us say that you made 3 commits A, B, C,
+and you have CS issues in A and in C.
+To make sure that is no longer the case, fixing each of these commits
+can be done like this:
+``git rebase --exec "vendor/bin/phpcbf && vendor/bin/phpcs" A^``
+That command will run phpcbf and then phpcs for each of your commits and
+will halt for A and C, but not for B because in the case of B they would
+exit with a zero status code. That will let you amend A, after which you
+can resume the rebase until you do the same for C. Here is how it would
+look like on A:
+
+    $ vendor/bin/phpcs      # check for issues phpcbf could not fix
+    $ git add -p            # commit whatever issues were fixed
+    $ git commit --amend    # change A
+    $ git rebase --continue # resume the rebase
+
+You should be able to apply the example above with any tool we use in
+our CI pipelines, such as PHPUnit, PHPStan or Psalm.
+
+``git rebase --interactive`` is a really powerful tool and we barely
+scratched the tip of the iceberg here. If you want to learn more about
+it, we recommend you watch `this talk from Pauline Vos <https://youtu.be/uI1V7771plw?t=814>`_
+
+Of course, if you want to craft good commits with good messages, you
+will have a hard time if the changeset you are describing does too many
+things. That might very well happen if you notice small things along the
+way that are unrelated to your PR, but too small to warrant a separate
+one. ``git add --patch`` or ``git add -p`` will be of invaluable help to
+commit things separately.
+On the contrary, there are commits that typically do not need to exist,
+such as commits that fix coding style or address minor review comments.
+Bear in mind that the git log is not only aimed at reviewers, but also
+at anyone who wants to understand some change you made. Do not distract
+them with cs fixes. Instead, try to produce a commit that contains your
+changes *and* the necessary fixes to pass coding standard checks.
+Also, it's best if all of your commits pass the build, because that
+makes them ``git bisect`` friendly, but it also means they are likely to
+be revertable independently from other commits in your PR. While being
+revertable is not particularly crucial to us, it can help you decide
+whether to squash or whether to split. For instance, it would not make
+sense to revert a commit documenting a feature without also reverting
+the code for that feature. That means there should be only once commit
+with both the code and the docs here.
+
 Rebasing on upstream changes
 ----------------------------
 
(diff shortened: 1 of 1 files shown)


source/contribute.rst
```diff
@@ -192,6 +195,70 @@ this in case you want to read more about this:
...
+being said, sometimes you end up with many commits and it's too late for
+that.
+
+If you are in the fairly simple case where you want squash all your
+commits into one, you can take the following steps to achieve that:
+
+1. ``git rebase -i doctrine/3.2.x``, assuming ``doctrine`` is a git remote
+   that points to the target repository, and not your fork, and assuming
+   the target branch is 3.2.x. If you're not sure what your remotes are,
+   run ``git remote -vvv``, there should be your fork and the
```
morozov, line 227: `git remote -v` will do the same.

source/contribute.rst
```diff
@@ -192,6 +195,70 @@ this in case you want to read more about this:
...
+hard time if the changeset you are describing does too many things. That
+might very well happen if you notice small things along the way that are
+unrelated to your PR, but too small to warrant a separate one. ``git add
+--patch`` or ``git add -p`` will be of invaluable help to commit things
+separately.
+On the contrary, there are commits that typically do not need to exist,
+such as commits that fix coding style or address minor review comments.
+Bear in mind that the git log is not only aimed at reviewers, but also
+at anyone who wants to understand some change you made. Do not distract
+them with cs fixes and try to make it look as if you produced perfect
```
morozov, line 211: It took me three times to read it to make sure I understand it correctly :-) It could be interpreted as: "Do not distract […] and [do not] try to make it look […]".

I'd break it up into two sentences: Do not distract […]. Instead, try to make it look […].

Additionally, the "make it look" to me seems to carry some negative connotation as if we're asking the contributor to fake something. Instead, it could be worded as "make a commit that contains the changes you wanted to make _and_ you had to make on top of those because of the code style, reviews, etc" (the wording is to be adjusted, this is just about the delivery).

source/contribute.rst
```diff
@@ -192,6 +195,70 @@ this in case you want to read more about this:
...
+code from the start. Also, it's best if all of your commits pass the
+build, because that makes them ``git bisect`` friendly, but it also
+means they are likely to be revertable independently from other commits
```
morozov, line 214: While it's true in theory, I can barely remember the moments when a single commit from a PR was reverted. I don't have a problem with keeping it but it doesn't seem essential. Bisect-friendly is very important, on the other hand.

source/contribute.rst
```diff
@@ -192,6 +195,70 @@ this in case you want to read more about this:
...
+
+If you are in the fairly simple case where you want squash all your
+commits into one, you can take the following steps to achieve that:
+
+1. ``git rebase -i doctrine/3.2.x``, assuming ``doctrine`` is a git remote
+   that points to the target repository, and not your fork, and assuming
+   the target branch is 3.2.x. If you're not sure what your remotes are,
+   run ``git remote -vvv``, there should be your fork and the
+   holy/reference/base/doctrine/whatever-you-call-it repository.
+2. A window will show up with many lines, replace ``pick`` with
```
morozov, line 229: Would it make more sense to point to some good tutorial on interactive rebasing? Maybe even instead listing the details here? Otherwise, if this info is not sufficient for a developer w/o the corresponding skills, the will be blocked.

source/contribute.rst
```diff
@@ -192,6 +195,70 @@ this in case you want to read more about this:
...
 - `A note about git commit messages <https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>`_
 
+To squash or not to squash
+--------------------------
+
+Of course, if you want to craft good commit messages, you will have a
+hard time if the changeset you are describing does too many things. That
+might very well happen if you notice small things along the way that are
+unrelated to your PR, but too small to warrant a separate one. ``git add
+--patch`` or ``git add -p`` will be of invaluable help to commit things
```
morozov, line 205: I think before going this deep, in the very first paragraph, we should give the contributor the right to not deal with all our perfectionism and just squash everything in one commit. Most of the developers making tiny patches will be happy right after this point and won't have to read the rest.

source/contribute.rst
```diff
@@ -192,6 +195,70 @@ this in case you want to read more about this:
...
+   forget the ``--force`` option otherwise git will try to merge both
+   things together.
+
+If you are in a more complex case where you would very much like to keep
+your commit separate, there are other solutions.
+To take a specific example, let us say that you made 3 commits A, B, C,
+and you have CS issues in A and in C.
+To make sure that is no longer the case, fixing each of these commits
+can be done like this:
+``git rebase --exec "vendor/bin/phpcbf && vendor/bin/phpcs" A^``
```
morozov, line 243: It's worth adding that depending on the case (if it's not CS but something else), one may have to run PHPStan, Psalm, and/or PHPUnit in a similar fashion.

source/contribute.rst
greg0ire: Maybe yeah, I think I have found something authoritative enough: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing
They use squash though, not fixup, so there is a risk that we end up with bad commit messages. Maybe let's experiment with that and see if it results in something like that or not…

source/contribute.rst
```diff
@@ -192,6 +195,69 @@ this in case you want to read more about this:
...
+will have a hard time if the changeset you are describing does too many
+things. That might very well happen if you notice small things along the
+way that are unrelated to your PR, but too small to warrant a separate
+one. ``git add --patch`` or ``git add -p`` will be of invaluable help to
+commit things separately.
+On the contrary, there are commits that typically do not need to exist,
+such as commits that fix coding style or address minor review comments.
+Bear in mind that the git log is not only aimed at reviewers, but also
+at anyone who wants to understand some change you made. Do not distract
+them with cs fixes. Instead, try to produce a commit that contains the
```
SenseException, line 247:
```suggestion
them with cs fixes. Instead, try to produce a commit that contains
```

SenseException (changes requested):

source/contribute.rst
SenseException: A new page could be created with more info about the different steps later if anything needs a deeper explanation. If the contribution entry page gets too big, it might discourage first time contributors.

source/contribute.rst
SenseException: > You should be able to apply the example above with any tool we use in
> our CI pipelines, such as PHPUnit, PHPStan or Psalm.

I assume this was added to resolve your mention of our other tools.

ostrolucky:
Like I mentioned on Slack, I don't like that people are amending commits. It fucks up reviews. Some contributors are completely changing the patches between amends. Hence I don't like this PR basically telling them to do that. Instead, I would welcome telling users to not to bother with squashing if their PR is simple. We can do that with simple click... That will cover most PRs and they can stop reading the section at that point. Rest will be still useful for more complex PRs.

source/contribute.rst
greg0ire: Yes, I added it because of @morozov 's comment

greg0ire:
> Or we can tell them not to squash if they are completely changing the commit and they think somebody is reviewing the PR, something like that?

Maybe we can recommend they use `git commit --fixup=HEAD` to make their intent clear?

SenseException:
Multiple "Fix CS" commits could be prevented with a `git commit --amend`, which might be easier to handle for some contributors who are not that used to git in their daily work. Of course this would only help to prevent multiple consecutive commits with the same scope of code changes and won't help when those redundant commits are already done.

ostrolucky:
Yeah not sure it will benefit us if we need to use squash button anyway. Also guide is already quite long. But I'll let you finish this however you want, I just wanted to mention the problem I see sometime, in case you might want to address that somehow.

greg0ire:
I think what would be best is my usual disclaimer:

> if you don't [squash], we'll try to remember to do it for you but it's best if you save us this trouble.

SenseException (approved):

morozov (approved):

PR #414, opened 2021-10-05
Split step into several steps

greg0ire:
It should help with understanding the workflow run.

### DIFF
--- a/.github/workflows/website-configs.yml
+++ b/.github/workflows/website-configs.yml
@@ -49,8 +49,17 @@ jobs:
       if: "steps.composer-cache.outputs.cache-hit != 'true'"
       run: "composer install --prefer-dist --no-progress --no-suggest --no-dev"
 
-    - name: "Prepare Webite files"
-      run: "bin/console --env=test sync-repositories && bin/console --env=test sync && bin/console --env=test clear-build-cache  && bin/console --env=test build-website-data"
+    - name: "Checkout repositories"
+      run: "bin/console --env=test sync-repositories"
+
+    - name: "Sync repositories"
+      run: "bin/console --env=test sync"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/website-configs.yml
```diff
@@ -49,8 +49,17 @@ jobs:
       if: "steps.composer-cache.outputs.cache-hit != 'true'"
       run: "composer install --prefer-dist --no-progress --no-suggest --no-dev"
 
-    - name: "Prepare Webite files"
-      run: "bin/console --env=test sync-repositories && bin/console --env=test sync && bin/console --env=test clear-build-cache  && bin/console --env=test build-website-data"
+    - name: "Checkout repositories"
+      run: "bin/console --env=test sync-repositories"
+
+    - name: "Sync repositories"
+      run: "bin/console --env=test sync"
```
greg0ire, line 56: @SenseException I initially wanted to fix the typo in "website", but then noticed this long line and decided it would be better to split it.

Then I noticed this step. I could not find a command named `sync`, and I think what's actually happening here is Symfony shortcut feature is allowing us to run the previous command (so `sync-repositories`) again. Is this really necessary?

.github/workflows/website-configs.yml
```diff
@@ -49,8 +49,17 @@ jobs:
...
+
+    - name: "Clearing build cache"
+      run: "bin/console --env=test clear-build-cache"
```
greg0ire, line 59: @SenseException likewise, is this really necessary? We're in the CI, so we are starting afresh, right?

SenseException:
website-configs.yml is obsolete and will be removed once the deployment is completely moved to GitHub Actions. The current staging branch is a wip and already removed this file.

Issue #388, opened 2021-02-15
Move deployment to GitHub Actions


SenseException:
The current deployment goes as the following, simplified, description:

* Change is pushed to the repository
* A server with a running cronjob realize a change happened and starts a website build
* After the website build on the server finished successfully, it deploys/pushes the generated static html to doctrine/doctrine-website-build-prod

In short: It fetches from GitHub and pushes it back to GitHub. Since GitHub Actions is available I think that the build-server should be replaced by GitHub Actions and we drop the server.

* We can build the website (and already do in CI) with GitHub Actions and the deployment to doctrine/doctrine-website-build-prod
* We don't need to update software of an always running and available server. We would have GitHub Actions workflows
* Updates of software like PHP should be easier with GitHub Actions
* GitHub Actions workflows are versioned while our current build server is not
* Builds in GitHub Actions are more visible and transparent in comparison. We can act faster on breaking builds (A first step of this was #351)

I don't know if there is a way to make a deployment happen on any documentation change of the other supported project-repositories, but I don't think we need that. A simple first step could be to deploy daily at the evening or at night. Maybe a deeper look into the deployment make some other ideas pop up for this.

PR #415, opened 2021-10-07
Introduce GitHub Actions deployment (prod)

SenseException:
This PR handles issue #388.

This workflow handles the deployment to the Doctrine website prod to move the deployment away from the build-server.

A daily deployment is planned as a cron to create the current documentation for all projects. I'm not sure if I should put prod and staging deployment into the same workflow file. They are mostly identical except the environment and the target branch.

What's left is the `event-participants` command that is supposed to save event participants and send emails to newly added participants, which are usually compared with entries in a database table. I'm not sure how highly frequented this part of the website is.


.github/workflows/deployment-prod.yml
```diff
@@ -0,0 +1,73 @@
...
+        restore-keys: |
+          ${{ runner.os }}-php-
+    - name: "Setup PHP Action"
+      uses: "shivammathur/setup-php@v2"
+      with:
+        php-version: "7.2"
+
+    - name: "Install dependencies"
+      if: "steps.composer-cache.outputs.cache-hit != 'true'"
+      run: "composer install --no-progress --no-suggest --no-dev"
```
greg0ire, line 50: You could use `ramsey/composer-install` here, it handles the cache part (so you could drop lines 35-42: https://github.com/doctrine/.github/blob/14d0040959e328daf3f0b34960ae84b0fc60bda5/workflow-templates/coding-standards.yml#L33-L36

.github/workflows/deployment.yml
```diff
@@ -1,4 +1,4 @@
-name: "deployment"
+name: "Deployment staging"
```
greg0ire, line 1: Maybe you could reuse workflows to eliminate duplication: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows ? It's a brand new feature.

.github/workflows/coding-standards.yml
```diff
@@ -24,15 +24,10 @@ jobs:
...
-        with:
-          path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
-          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
-
       - name: "Install dependencies with Composer"
-        run: "composer install --no-interaction --no-progress --no-suggest"
+        uses: "ramsey/composer-install@v1"
+        with:
+          composer-options: "--no-interaction --no-progress --no-suggest"
```
greg0ire, line 30: In fact, for this workflow, I think you can reuse the one from the `.github` repository :slightly_smiling_face:

PR #430, opened 2021-12-02
Update root-folders only

malarzm:
Fixes https://github.com/doctrine/mongodb-odm/issues/2099

Previously `reference/reference-mapping` was being changed to `/reference//reference-mapping`.

### DIFF
--- a/lib/Docs/RST/RSTCopier.php
+++ b/lib/Docs/RST/RSTCopier.php
@@ -168,7 +168,7 @@ SIDEBAR;
             $sidebar = $this->rstFileRepository->getFileContents($sidebarPath);
 
             // sidebar.rst paths were wrong. Remove this once sidebar.rst is updated everywhere
-            $sidebar = preg_replace('/(reference|tutorials|reference|cookbook)/', '/$1', $sidebar);
+            $sidebar = preg_replace('/^([ \t]*)(reference|tutorials|cookbook)/m', '$1/$2', $sidebar);
 
             return $sidebar;
         }


malarzm:
@SenseException could you please retry the failing action when you have a chance? Seems random but I have no power here :)

SenseException:
I get a bunch of errors like

```
/!\ Error while processing "toctree" directive in "reference/reference-mapping": The "/websitepath/cache/docs/doctrine-mongodb-odm/en/latest/reference/tutorials/" directory does not exist.
/!\ Error while processing "toctree" directive in "reference/reference-mapping": The "/websitepath/cache/docs/doctrine-mongodb-odm/en/latest/reference/cookbook/" directory does not exist.
```

`websitepath` is the path of the website project.

malarzm:
That's weird, https://raw.githubusercontent.com/doctrine/mongodb-odm/2.3.x/docs/en/reference/reference-mapping.rst has no `toctree` in it 🤔

malarzm:
Ah, I believe the sidebar gets inlined to each file and either the file wasn't picked up before, or my regex is somehow wrong

malarzm:
The regex should be better now: I forgot the `m` flag. Also changed `\s` to explicit spaces and tabs so it won't match new lines now

SenseException (approved):
Some links appeared again that were missing before.

Great work. :+1:

malarzm:
@SenseException is there any way to rebuild ODM's documentation? Links are still missing on https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/index.html

SenseException:
The build is currently not working. The node-sass component breaks despite no changes were introduced in the build process. I started to introduce some changes in a branch, but it seems my efforts didn't bring positive results yet.

PR #432, opened 2022-01-02
Update PHP to (interim) 7.4

SenseException:
I actually wanted to include some updates for the frontend too, but some image size issues in the search-box after the webpack build I wasn't able to fix yet made me reduce the scope to the backend for this PR.

I chose to update to PHP 7.4 first and later to PHP 8.x in case some of our Doctrine dependencies need an update too. Also because some Composer dependencies are going to be major versions.

Frontend will be handled in a separate PR.


.github/workflows/php7.2.yml
```diff
@@ -4,7 +4,7 @@ on: ["push", "pull_request"]
 
 env:
   doctrine_website_github_http_token: "${{ github.token }}"
-  PHP_VERSION: "7.2"
+  PHP_VERSION: "7.4"
```
greg0ire, line 7: Hmmmm… the name of that file contains the php version

SenseException: Yeah, a stupid decision. I'll rename it to `continuous-integration.yml`.

greg0ire (approved):

Issue #200, opened 2018-08-10
Deprecation summary


jwage:
The point of this GitHub issue is to discuss in a more organized way the different ways we can handle deprecations in Doctrine.

The policy of Doctrine for a very long time has been to use `@deprecated` and only very recently did we start to introduce usages of `@trigger_error` which has led to this discussion of how they should be used and in which use cases.

## Using `trigger_error()` with `@` and `phpunit-bridge`.

- It is always called in production.
- Causes bug reports to us that aren't bugs and we have to explain it is a deprecation error and they should handle it properly in their build with something like phpunit bridge.
- People who don't use `phpunit-bridge` would have to depend on the CHANGELOG to know what has changed.
- It won't report to you calls to deprecated code if your test suite does not execute that code path or the code that is calling the deprecated code is dead.
- `phpunit-bridge` reports deprecated method calls that your project indirectly calls. See 3 and 4 in this comment https://github.com/doctrine/common/pull/849#issuecomment-407053196

## Using `@deprecated` and `phpstan-deprecation-rules`.

- Means other major dependents won't be automatically notified about deprecations if they aren't using `phpstan-deprecation-rules`.
- It seems like the `phpunit-bridge` can handle SOME `@deprecated` scenarios, but it is not clear to me the extent of this functionality.
- If we immediately switch to this, any dependents of Doctrine won't be automatically notified of deprecations if they aren't using `phpstan-deprecation-rules`. However, this is the same as if we are using `@trigger_error` and someone isn't using phpunit-bridge.
- People who don't use `phpstan-deprecation-rules` would have to depend on the CHANGELOG to know what has changed.
- `phpstan-deprecation-rules` will tell you when you call something that is deprecated, even if your test suite doesn't execute the code or the code is dead. `phpstan` will also notify you of the dead code that could be removed which gets rid of the call to the deprecated code.
- What happens in the scenario when a Doctrine column type that’s deprecated is referenced in XML?
  - @lcobucci is chatting with phpstan folks to see if things can be improved to handle the above two "problems"
- What happens if a users project uses Twig and they call a Twig function that calls something deprecated in Doctrine?
  - PHPStan will see that you have a method that is calling a deprecated method. Even if that parent method is never called in your test suite/build process.
- What happens when you have a conditional deprecation? How does this work with `phpstan-deprecation-rules`? e.g.
  - Here we would use `trigger_error()` without the `@` and suggest to people to appropriately configure PHP with `error_reporting()`

```php
<?php

[...]

PR #448, opened 2022-04-24
Recommend Doctrine Deprecations over trigger_error

greg0ire:
This is representative of what we have been doing for a while.

### DIFF
--- a/source/policies/deprecation.rst
+++ b/source/policies/deprecation.rst
@@ -55,44 +55,12 @@ Then, if your code directly depends on ``doctrine/orm`` which provides the
         }
     }
 
-Using ``trigger_error()`` and ``phpunit-bridge``
-------------------------------------------------
+Using ``doctrine/deprecations``
+-------------------------------
 
-In some cases, you may need to conditionally deprecate functionality. In these cases, it is required
-to use the ``trigger_error()`` function and a tool like `PHPUnit Bridge`_ which implements a custom
-error handler for you to use to collect and report these warnings. **This method is not recommended
-and Doctrine will try to avoid using it if possible.**
+In some cases, you may need to conditionally deprecate functionality. In
+these cases, it is required to use the `Doctrine deprecations`_ library.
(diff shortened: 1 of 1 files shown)


source/policies/deprecation.rst
```diff
@@ -55,44 +55,12 @@ Then, if your code directly depends on ``doctrine/orm`` which provides the
...
 
-Using ``trigger_error()`` and ``phpunit-bridge``
-------------------------------------------------
+Using ``doctrine/deprecations``
+-------------------------------
 
-In some cases, you may need to conditionally deprecate functionality. In these cases, it is required
-to use the ``trigger_error()`` function and a tool like `PHPUnit Bridge`_ which implements a custom
-error handler for you to use to collect and report these warnings. **This method is not recommended
-and Doctrine will try to avoid using it if possible.**
```
greg0ire, line 64: Note: here I'm removing the sentence about this method being not recommended, because I don't see any issue with it when using a `psr/logger`, but I'm afraid that might be a personal opinion. I think at least @Ocramius has things to say on the matter, but it's unclear to me what. See https://github.com/doctrine/persistence/pull/281#issuecomment-1101638803

I read https://github.com/doctrine/doctrine-website/issues/200, but it doesn't mention `doctrine/deprecations`.

Personally, I'm not sure trying hard to go the `@deprecated` way every time is the way to go: the implication is that you have to come up with new names for classes/methods/etc. in favor of which to deprecate existing counterparts. This could result in an awkward API because of bad naming.

greg0ire:
@derrabus for some reason you're not in the list, but please review.

greg0ire:
Also, @malarzm :)

greg0ire:
@derrabus @morozov @malarzm for some reason you're not in the list, but please review

Ocramius (approved):

malarzm (approved):

derrabus (approved):

PR #474, opened 2023-01-25
Remove Google Translate

### LINKED ISSUE #408: High CPU usage when idle
The CPU usage is high even when idle, between 15 and 30% on my laptop.

The source seem to be a loader (a SVG with a css animation) injected by the google translate javascript.
The animation triggers repaint events every second or so.

The repaint events can be seen using the profiler (just let it run 5 seconds without doing anything) and the following snippet to find the source:

```
window.onanimationiteration = console.log;
```

The real issue is on google side - why keep an animated spinner forever ? It doesn't even seem to be displayed on the page - but I can't find where such issues must be filed.

Also:
Setting display: none on the container of the animation stop the repaint events in Chrome.

Screenshot from 2021-07-21 13-59-37

Screenshot from 2021-07-21 13-57-16

Screenshot from 2021-07-21 13-50-15

### LINKED ISSUE #422: Suggestion: Remove Google translation bar
I'm suggesting to remove the Google translation bar that was introduced in https://github.com/doctrine/doctrine-website/pull/141

Why?
1. Since it's loaded with JavaScript, it's causing a so-called Cumulative Layout Shift (CLS, see https://web.dev/cls/): After everything is on screen, the translation bar appears, moving everything down.
2. In that toolbar I've clicked the "Deactivate for English" button *at least* ;-) 100 times, but it keeps reappearing every time...

greg0ire:
There is doubt on how useful it is, and there is no doubt on how annoying it is.

Closes #422
Fixes #408

### DIFF
--- a/source/index.html
+++ b/source/index.html
@@ -1,7 +1,7 @@
 {% block content %}
     <div class="row">
         <div class="col-12 col-md-{{ featuredPartner ? 6 : 12 }} mb-4">
-            <h1 class="display-4 notranslate">{{ site.title }}</h1>
+            <h1 class="display-4">{{ site.title }}</h1>
 
             <p class="lead">The Doctrine Project is the home to several PHP libraries primarily
             focused on database storage and object mapping. The core projects are the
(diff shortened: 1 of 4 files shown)


SenseException (approved):

SenseException:
Thanks @greg0ire

Issue #475, opened 2023-01-27
Update website to PHP 8.2


SenseException:
After the in-between steps of #435 and #432 and updating our own outdated Composer dependencies that are used by the website, we're at a point where I can now start the update for the website itself.

The raw steps I planned (not necessarily in that order) are

* Update PHP, dependencies and code of the website
* Update the Frontend dependencies and code of the website (not really PHP related but needs to be done)
* Maintain the code of the website like
  * dropping unused code
  * dropping dependencies (ununsed ones and used ones like jQuery)
  * refactorings/rewrites
  * handle rst-parser separately because of it's still growing work in progress

For now I focus on the first two points of this issue which are the main tasks of the update. I'll already began with PHP.

SenseException:
PHP Update is done and also some of the JS dependencies were updated. Some unused code was dropped too, but there might be more of it that needs to be removed.

SenseException:
Update's done so far.

PR #476, opened 2023-01-31
Base update to PHP 8.2

SenseException:
This PR is for the plain update to PHP 8.2 with its composer dependencies. Refactorings and rewrites aren't included in this step.


SenseException:
The update is pretty much close to the end, but generating the docs end in multiple errors based on the outdated rst format of the doc-files, especially 

* the tables in dbal (Cannot find column for index "0")
* doctrine-data-fixtures (Failed building file "source/projects/doctrine-data-fixtures/en/1.6/how-to/sharing-objects-between-fixtures.html" with error "Unexpected token "name" of value "if" ("end of statement block" expected))

I don't know how the docs will look like either until I manage a docs-build without exceptions.

greg0ire:
> doctrine-data-fixtures (Failed building file "source/projects/doctrine-data-fixtures/en/1.6/how-to/sharing-objects-between-fixtures.html" with error "Unexpected token "name" of value "if" ("end of statement block" expected))

Might be because of the linebreak right in the middle of a link? There's only one occurrence of `if` in [this document](https://raw.githubusercontent.com/doctrine/data-fixtures/1.6.x/docs/en/how-to/sharing-objects-between-fixtures.rst)

greg0ire:
Cc @linawolf in case you this might be a bug in the parser, or in case linebreaks in the middle of a link are invalid.

SenseException:
When I remove doctrine/data-fixtures from the build then the next project is affected by the same error. In my case`source/projects/doctrine-orm-module/en/5.0/index.html`.

greg0ire:
:thinking: no occurrence of `if` in that one

linawolf:
There needs to be a new line after `.. code-block:: php` otherwise the codeblock is invalid

greg0ire:
I created https://github.com/doctrine/data-fixtures/pull/428 to address this on the data-fixtures repo.

SenseException:
> Failed building file "source/projects/doctrine-data-fixtures/en/1.6/how-to/sharing-objects-between-fixtures.html" with error "Unexpected token "name" of value "if" ("end of statement block" expected)

I'm not sure if this error is connected to the rst-parser. The html file already exists and contains html+twig code, so the rst got parsed.

[sharing-objects-between-fixtures.txt](https://github.com/doctrine/doctrine-website/files/10771877/sharing-objects-between-fixtures.txt)

greg0ire:
Should you maybe add `-vvv` somewhere to get a full stack trace?
Here? https://github.com/doctrine/doctrine-website/blob/a13875d79e85af9f89d2507defe092ebba98f2fa/.github/workflows/deployment.yml#L65

SenseException:
The workflow breaks earlier because of the dbal table I mentioned in a comment earlier, but I already got myself a stack-trace on my local machine. I'd expect a Twig error, but I haven't pinned it down to the actual cause.

This file contains the complete output of one of my test runs. (I skipped the Exceptions of the dbal tables)
[build.txt](https://github.com/doctrine/doctrine-website/files/10771954/build.txt)

greg0ire:
Looking at that stack trace, I think it might be about this line: https://github.com/doctrine/doctrine-website/blob/a13875d79e85af9f89d2507defe092ebba98f2fa/templates/layouts/documentation.html.twig#L119

Of maybe this one: https://github.com/doctrine/doctrine-website/blob/a13875d79e85af9f89d2507defe092ebba98f2fa/templates/layouts/documentation.html.twig#L139

greg0ire:
Ah yes it checks out: https://github.com/twigphp/Twig/blob/78c18ad58e48d3fff2f7812bb60b6176ee52689a/CHANGELOG#L156

greg0ire:
The failure seems to relate to https://github.com/doctrine/dbal/pull/5197

greg0ire:
But if we had to reuse this commit on lower branches, well there are a lot of branches to backport it to: https://github.com/doctrine/dbal/blob/3.6.x/.doctrine-project.json

I don't understand why this wasn't an issue sooner :thinking:

SenseException:
Yes, it's the newer versions of the rst-parser. It's breaking the website build in the latest 0.5 release and the 0.6 branch. I've created an issue some time about the behavior of one of the errors, but I don't find it anymore and I'm not sure if it's the same issue again. Once my workstation is ready to use, I can resume my work.

SenseException:
The build-breaking exception happens on dbal only 12 times

```
Cannot find column for index "0"
#0 doctrine-website/vendor/doctrine/rst-parser/lib/Nodes/TableNode.php(188): Doctrine\RST\Nodes\TableNode->compilePrettyTable()
#1 doctrine-website/vendor/doctrine/rst-parser/lib/Nodes/TableNode.php(146): Doctrine\RST\Nodes\TableNode->compile()
#2 doctrine-website/vendor/doctrine/rst-parser/lib/Parser/DocumentParser.php(587): Doctrine\RST\Nodes\TableNode->finalize()
#3 doctrine-website/vendor/doctrine/rst-parser/lib/Parser/DocumentParser.php(378): Doctrine\RST\Parser\DocumentParser->flush()
#4 doctrine-website/vendor/doctrine/rst-parser/lib/Parser/DocumentParser.php(216): Doctrine\RST\Parser\DocumentParser->parseLine()
#5 doctrine-website/vendor/doctrine/rst-parser/lib/Parser/DocumentParser.php(153): Doctrine\RST\Parser\DocumentParser->parseLines()
#6 doctrine-website/vendor/doctrine/rst-parser/lib/Parser.php(179): Doctrine\RST\Parser\DocumentParser->parse()
#7 doctrine-website/vendor/doctrine/rst-parser/lib/Parser.php(163): Doctrine\RST\Parser->parseLocal()
#8 doctrine-website/vendor/doctrine/rst-parser/lib/Parser.php(205): Doctrine\RST\Parser->parse()
#9 doctrine-website/vendor/doctrine/rst-parser/lib/Builder/ParseQueueProcessor.php(81): Doctrine\RST\Parser->parseFile()
#10 doctrine-website/vendor/doctrine/rst-parser/lib/Builder/ParseQueueProcessor.php(65): Doctrine\RST\Builder\ParseQueueProcessor->processFile()
#11 doctrine-website/vendor/doctrine/rst-parser/lib/Builder.php(214): Doctrine\RST\Builder\ParseQueueProcessor->process()
#12 doctrine-website/vendor/doctrine/rst-parser/lib/Builder.php(151): Doctrine\RST\Builder->parse()
#13 doctrine-website/lib/Docs/RST/RSTBuilder.php(57): Doctrine\RST\Builder->build()
#14 doctrine-website/lib/Docs/RST/RSTBuilder.php(36): Doctrine\Website\Docs\RST\RSTBuilder->buildRst()
#15 doctrine-website/lib/Docs/BuildDocs.php(92): Doctrine\Website\Docs\RST\RSTBuilder->buildRSTDocs()
#16 doctrine-website/lib/Commands/BuildDocsCommand.php(62): Doctrine\Website\Docs\BuildDocs->build()
#17 doctrine-website/vendor/symfony/console/Command/Command.php(326): Doctrine\Website\Commands\BuildDocsCommand->execute()
#18 doctrine-website/lib/Commands/BuildAllCommand.php(101): Symfony\Component\Console\Command\Command->run()
#19 doctrine-website/lib/Commands/BuildAllCommand.php(85): Doctrine\Website\Commands\BuildAllCommand->runCommand()
#20 doctrine-website/vendor/symfony/console/Command/Command.php(326): Doctrine\Website\Commands\BuildAllCommand->execute()
#21 doctrine-website/vendor/symfony/console/Application.php(1063): Symfony\Component\Console\Command\Command->run()
#22 doctrine-website/vendor/symfony/console/Application.php(320): Symfony\Component\Console\Application->doRunCommand()
#23 doctrine-website/vendor/symfony/console/Application.php(174): Symfony\Component\Console\Application->doRun()
#24 doctrine-website/lib/Application.php(121): Symfony\Component\Console\Application->run()
#25 doctrine-website/bin/console(21): Doctrine\Website\Application->run()
```

SenseException:
Looks like ORM and DBAL are the only ones that cause the build to break now. If I skip/ignore these, then the build runs to the end and I can see the website with its docs (minus the versions that cause the problems)

ORM
-------

https://github.com/doctrine/rst-parser/blob/0.5.x/lib/Parser/LineDataParser.php#L177-L179

These lines are the ones that are involved in the ORM build failure "Warning: Undefined array key 0". `definitionNodes` is an empty array at this point starting with 2.12.x (and probably lower). Considering the content that was parsed at the time I assume that strategy-cookbook-introduction.rst was parsed at the point of where the build breaks. I can't tell yet if the empty array could be correct, because the only difference is, that the paragraph that starts with "As you can see, we have a method setBlockEntity which" is bold.

DBAL
--------

The newer versions work fine. 3.3 is the last version that gets its docs generated, but 3.2 is the first in line that causes the build to break. The trace of the exception in my previous comment reduces the source of this error to the types-table. A diff between 3.3 and 3.2 shows an indent of the table and reference fixes as the only changes.

greg0ire:
> As you can see, we have a method setBlockEntity which

With 2.12.x, that sentence is preceded with a seemingly empty line. Seemingly, because it in fact has trailing whitespace (3 spaces). It would be interesting to push a commit removing that trailing whitespace to 2.12.x and see what happens.

SenseException:
I came to the same conclusion, because `$definitionListTerm['definition']` contains a `\n` . Maybe that behavior can be reproduced in a rst-parser test.

greg0ire:
Yes and I think the fix will be to change line 312 of `DocumentParser.php` to `if (trim($this->lines->getNextLine()) !== '' && $this->lineChecker->isIndented($this->lines->getNextLine())) {`

greg0ire:
Ah hang on, if I copy paste the file literally, I do reproduce the issue. I'll try reducing it to what's strictly necessary.

greg0ire:
Oh and my fix works. I must have copy/pasted it wrong.

greg0ire:
The desired behavior depends on the definition of "blank line". Can a blank line contain whitespace? @linawolf , surely you must know? With https://github.com/wouterj/sphinx-tests, I see that Sphinx parses 

```
test
 
test
```
as 2 paragraphs, so I would say the second line is considered blank.

greg0ire:
Pushed a new commit. Now we have to deal with the DBAL docs.

greg0ire:
I can happily confirm that fixing this table addresses the issue :)

greg0ire:
I don't know how to fix it from  DBAL though: that would imply creating new tags I think. The next best thing might be to skip the problematic row.

SenseException:
I found my comment again in https://github.com/doctrine/dbal/issues/3338#issuecomment-657255122 that shows what you describe. I thought I created an issue about it which is why I couldn't find it at first.

> The next best thing might be to skip the problematic row.

Changing the older versions shouldn't be the solution, so I agree. We just postponed the dbal issue :smiley: 
We should also consider to drop older versions in the future, but that's a topic we should discuss another time..

greg0ire:
Just tried sphinx on that table, and it does not seem to find anything wrong with it. It produces this:



Here is the corresponding html:

```html
<tr class="row-odd">
    <td rowspan="6"><p><strong>binary</strong> [2]  [6]</p></td>
    <td rowspan="6"><p><code class="docutils literal notranslate"><span class="pre">resource</span></code></p></td>
    <td><p><strong>MySQL</strong></p></td>
    <td rowspan="4"><p><em>all</em></p>
        <p><em>all</em></p>
    </td>
    <td rowspan="2"><p><code class="docutils literal notranslate"><span class="pre">VARBINARY(n)</span></code> [3]</p></td>
</tr>
<tr class="row-even"><td rowspan="2"><p><strong>SQL Server</strong></p></td></tr>
<tr class="row-odd"><td rowspan="2"><p><code class="docutils literal notranslate"><span class="pre">BINARY(n)</span></code> [4] <code class="docutils literal notranslate"><span class="pre">RAW(n)</span></code></p></td></tr>
<tr class="row-even"><td><p><strong>Oracle</strong></p></td></tr>
<tr class="row-odd"><td><p><strong>PostgreSQL</strong></p></td>
    <td><p><em>all</em></p></td>
    <td><p><code class="docutils literal notranslate"><span class="pre">BYTEA</span></code> [15]</p></td>
</tr>
<tr class="row-even"><td><p><strong>SQLite</strong></p></td>
    <td><p><em>all</em></p></td>
    <td><p><code class="docutils literal notranslate"><span class="pre">BLOB</span></code> [15]</p></td>
</tr>
```

This means that it might be possible to fix doctrine/rst-parser to do the same… in theory. The table is nonsensical and malformed I think, but still renderable.

SenseException:
`rowspan="4"` for the cell that contains "all all". 3 cells on the left and two on the right.

greg0ire:
I'm unfamiliar with it and would need time and to be fully rested to even begin to understand how it works :sweat:

SenseException:
Same here. 

>  I'll take a look what options the current website code offers for such a problem.

Looks like this can't be handled on the website side and has to be tackled in the rst-parser.

derrabus:
Can we split non-blocking parts out of this PR, e.g. changes like #492 or #493?

SenseException:
@derrabus This is mostly an update of PHP and PHP dependencies with follow-up PRs after this one (see #475). If some PRs can be done now that won't make rebases too hard it should be fine.

derrabus:
I'll see what I can split out. If someone gives me write access to the repo, I can also keep this PR rebased.

SenseException:
The JS parts were supposed to be done separately from PHP anyway, so thanks for those now rebased PRs :+1:

lib/Event/TableIncompatibility.php
SenseException, line 1: I'm not really happy doing this workaround. It seems like the dbal table issue of the older versions can't be fixed with the current TableNode code in a simple manner (https://github.com/doctrine/rst-parser/pull/273#issuecomment-1688862762).

greg0ire: Once this is merged, it should be possible to work on ditching `doctrine/rst-parser` in favor of `phpDocumentor/guides`. Hopefully we can remove that workaround then.

greg0ire:
Congrats on getting a green build!

greg0ire (approved):

flagbird:
Before converting this draft to a PR, and therefore before merging this, We have to test the website on staging. I did a deployment yesterday, but for yet unknown reasons, some versions are missing in the dbal docs. There are probably other projects affected too. This needs to be checked if this issue is connected to the updates of this branch.

SenseException:
Before converting this draft to a PR, and therefore before merging this, We have to test the website on staging. I did a deployment yesterday, but for yet unknown reasons, some versions are missing in the dbal docs. There are probably other projects affected too. This needs to be checked if this issue is connected to the updates of this branch.

lib/Event/TableIncompatibility.php
SenseException: There are also some other things I want to fix on the website that I planned for after this update. Those will be much smaller PRs :sweat_smile:

SenseException:
Let's do this :muscle:

greg0ire:
Thanks and congrats @SenseException !!!

derrabus:
Great work!

PR #479, opened 2023-04-17
Update `2022-01-11-orm-2.11.md` to be clear about some limitations over "Support for Readonly Properties"

n4huel:
The support for read only properties only works fine for entities which use the "None" strategy for identifier generator. When the entity use other strategy, like the suggested in the documentation, an exception is thrown when you try to remove the entity.

See: 
- https://github.com/doctrine/orm/issues/10032
- https://github.com/doctrine/orm/blob/2.14.x/lib/Doctrine/ORM/UnitOfWork.php#L1271-L1273
- https://github.com/doctrine/orm/blob/2.14.x/lib/Doctrine/ORM/Mapping/ReflectionReadonlyProperty.php#L46
- https://www.php.net/manual/en/language.oop5.properties.php#language.oop5.properties.readonly-properties
- https://3v4l.org/FrhLN

### DIFF
--- a/source/blog/2022-01-11-orm-2.11.md
+++ b/source/blog/2022-01-11-orm-2.11.md
@@ -88,20 +88,24 @@ Another PHP 8.1 feature is the new readonly keyword that prevents the value of
 a property to be written again after it has been initialized in the constructor
 of an object.
 
-With ORM 2.11 the support now works as you would expect with no additional
-mapping options necessary:
+With ORM 2.11 the support now works only for entities which use the "NONE"
+strategy for identifier generator:
 
 ```php
 #[Entity, Table(name: 'author')]
 class Author
 {
-    #[Column, Id, GeneratedValue]
+    #[Column, Id, GeneratedValue(strategy: 'NONE')]
     private readonly int $id;
 
     #[Column]
     private readonly string $name;
 }
 ```
+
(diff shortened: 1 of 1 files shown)


SenseException:
Hi @n4huel, and thank you for your contribution. This is a blog post from January 2022 about ORM 2.11. I don't know if an old blog post is supposed to be changed.

@beberlei how do you see it? should we adapt an old blog post? I'm asking you because this is one you've created.

n4huel:
Thanks you @SenseException, your point is accurate. Maybe it should include some note/disclaimer update as "errata" since the feature isn't really supported.

stof:
"it will not work" is also limited to the id field. Other fields can still be readonly.

Novynn:
Doctrine ORM still does not support readonly ids due to this problem, and the blog post is the first thing that appears in a search.

vuryss:
Just found another bug with the readonly IDs and new PHP 8.4 ghost objects - if the ID is readonly of a related entity - when it tries to create a proxy for that related entity - the proxy is immediately initialized from the ReadonlyAccessor - it calls the reflection property->isInitialized on the ID and this triggers the initialization in the foreach right after the proxy generation. Here is the stacktrace from inside of the ghost object function to visualize which flow triggered the initialization.

image

PR #487, opened 2023-07-04
Avoid nesting blocks in non-capturing blocks

greg0ire:
This does not work with Twig 2.

See https://github.com/twigphp/Twig/pull/2687

Should help with #476


SenseException (approved):
Yes, it should. I haven't looked into the Twig templates for `^3.5` that much yet. Thanks.

PR #508, opened 2023-08-24
JS GitHub workflows

SenseException:


greg0ire:
Should [`paths`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) also be specified?

SenseException:
That would look like in the recent commit.

.github/workflows/js-analysis.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  push:
+    paths:
+      - '**.js'
+      - '**.scss'
+      - 'package.json'
+      - 'yarn.lock'
+      - 'jest.config.js'
+      - 'webpack.config.js'
+      - '.github/workflows/js-tests.yml'
```
greg0ire, line 12: :thinking: why would we need to run that other workflow when changing this one and vice versa?

.github/workflows/js-analysis.yml
```diff
@@ -0,0 +1,45 @@
...
+
+on:
+  push:
+    paths:
+      - '**.js'
+      - '**.scss'
+      - 'package.json'
+      - 'yarn.lock'
+      - 'jest.config.js'
+      - 'webpack.config.js'
```
greg0ire, line 11: Are these 2 lines not redundant with the first one?

derrabus (approved):

.github/workflows/js-analysis.yml
```diff
@@ -0,0 +1,39 @@
...
+      - 'yarn.lock'
+      - '.github/workflows/js-analysis.yml'
+
+jobs:
+  code-analysis:
+
+    runs-on: "ubuntu-latest"
+
+    steps:
+    - uses: "actions/checkout@v2"
```
greg0ire, line 25:
```suggestion
    - uses: "actions/checkout@v3"
```

.github/workflows/js-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      - 'yarn.lock'
+      - '.github/workflows/js-tests.yml'
+
+jobs:
+  js-tests:
+
+    runs-on: "ubuntu-latest"
+
+    steps:
+    - uses: "actions/checkout@v2"
```
greg0ire, line 25:
```suggestion
    - uses: "actions/checkout@v3"
```

greg0ire (approved):

greg0ire:
Approving in case you used "v2" in order to be consistent with other workflows, and plan to switch all workflows to v3 in another PR.

SenseException:
> Approving in case you used "v2" in order to be consistent with other workflows, and plan to switch all workflows to v3 in another PR.

Revisiting the workflows wasn't at the top of my todo list, but I can do at least update versions in a new PR in one of my next PRs.

PR #524, opened 2023-09-12
Use original CacheClearer class in test

SenseException:
The partial was removed to make `CacheClearer` test results more expressive. I wonder if I should use `Filesystem` directly instead of a mock because the test setup is already using the functionality of that class.


derrabus:
> I wonder if I should use `Filesystem` directly instead of a mock

Yes, mocking the `Filesystem` class is rarely a good idea. Better use a temp directory.

derrabus (approved):

PR #525, opened 2023-09-12
Increase DataBuilder code coverage

SenseException:
One of the tests is causing a PHPUnit warning because one of the tests forces a `false` return value from the function `file_get_contents()`, which is on purpose. Any ideas how I can tell that test that this is expected?

### DIFF
--- a/tests/DataBuilder/WebsiteDataReaderTest.php
+++ b/tests/DataBuilder/WebsiteDataReaderTest.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Tests\DataBuilder;
+
+use Doctrine\Website\DataBuilder\WebsiteData;
+use Doctrine\Website\DataBuilder\WebsiteDataReader;
+use org\bovigo\vfs\vfsStream;
+use PHPUnit\Framework\TestCase;
+use RuntimeException;
+
+class WebsiteDataReaderTest extends TestCase
+{
+    private WebsiteDataReader $websiteDataReader;
+
+    public function testRead(): void
+    {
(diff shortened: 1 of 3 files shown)


tests/DataBuilder/WebsiteDataReaderTest.php
```diff
@@ -0,0 +1,66 @@
...
+    public function testReadInvalidJson(): void
+    {
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Could not load JSON from file vfs://cache/data/invalid.json');
+
+        $this->websiteDataReader->read('invalid');
+    }
+
+    /**
+     * A PHPUnit warning is expected here because of file_get_contents not being able to open a file.
```
greg0ire, line 43: When I run phpunit, it tells me that this is actually a PHP warning:

```
1 test triggered 1 PHP warning:

1) /home/gregoire/dev/doctrine-website/lib/DataBuilder/WebsiteDataReader.php:33
file_get_contents(vfs://cache/data/im-a-directory-stupid.json): Failed to open stream: "org\bovigo\vfs\vfsStreamWrapper::stream_open" call failed

Triggered by:

* Doctrine\Website\Tests\DataBuilder\WebsiteDataReaderTest::testReadUnableToReadFileContent
  /home/gregoire/dev/doctrine-website/tests/DataBuilder/WebsiteDataReaderTest.php:45

```

SenseException: It's vfsStream that doesn't like it.  I've removed this test for now because it's not an important one. Maybe I'll use the real filesystem like in other tests for this case.

greg0ire (approved):

PR #529, opened 2023-09-18
Increase code coverage for Projects namespace

SenseException:


lib/Projects/GetProjectPackagistData.php
```diff
@@ -12,15 +12,19 @@ class GetProjectPackagistData
 {
     private const PACKAGIST_URL_FORMAT = 'https://packagist.org/packages/%s.json';
```
greg0ire, line 13: This constant is no longer in use I think.

tests/Projects/ProjectGitSyncerTest.php
```diff
@@ -111,4 +126,19 @@ public function testCheckoutBranch(): void
...
+    {
+        self::assertTrue($this->projectGitSyncer->isRepositoryInitialized('orm'));
+        self::assertFalse($this->projectGitSyncer->isRepositoryInitialized('foo'));
+    }
+
+    public function testCheckoutTag(): void
+    {
+        $this->processFactory->expects(self::once())
+            ->method('run')
+            ->with('cd \'vfs://projects/example-project\' && git clean -xdf && git checkout tags/\'1.0.0\'');
```
greg0ire, line 140: I would minimize the escaping here

```suggestion
            ->with("cd 'vfs://projects/example-project' && git clean -xdf && git checkout tags/'1.0.0'");
```

lib/Projects/GetProjectPackagistData.php
SenseException: It's used in the constructor as a default value. I made it injectable to be able to test the class. I can remove the default value in the constructor and put the url into the service definitions.

greg0ire (approved):

PR #532, opened 2023-09-26
Use final and readonly where possible

derrabus:
Finalizing classes/methods or flagging properties as readonly helps a reader understand how a class is used and what impact a code change may have. I propose to make `final` and `readonly` the default in this codebase.

### DIFF
--- a/lib/Application.php
+++ b/lib/Application.php
@@ -28,7 +28,7 @@ use function is_string;
 use function realpath;
 use function sprintf;
 
-class Application
+final readonly class Application
 {
     public const ENV_PROD    = 'prod';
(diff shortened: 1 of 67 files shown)


SenseException:
Why are some classes "just" `/** @final */`?

derrabus:
Because we create mocks for those classes. 😢

SenseException (approved):
I'm currently working on increasing the test coverage. I'll come back to the final-annotation in case a mock is serviceable. I don't want to use `nunomaduro/mock-final-classes` on new tests.

derrabus:
Sure, whenever you need to create a mock, feel free to un-finalize classes.

PR #534, opened 2023-10-03
Set timezone to America/New_York in phpunit

SenseException:
The website is working with America/New_York and so must the tests.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -13,6 +13,10 @@
         </testsuite>
     </testsuites>
 
+    <php>
+        <ini name="date.timezone" value="America/New_York"/>
+    </php>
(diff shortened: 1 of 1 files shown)


derrabus:
Can you elaborate? This setting feels a bit arbitrary.

SenseException:
The website has a timezone set in https://github.com/doctrine/doctrine-website/blob/master/lib/Application.php#L111 and tests involving DateTimeImmutable or DateTime in objects will fail due to differences when I have to provide date + time strings that will create a DateTimeImmutable instance during a running test. 

While writing tests for the existing code I don't want to change too much of the original website code.

derrabus (approved):

PR #538, opened 2023-10-11
ORM Beta 1, DBAL 4 RC 1 and future plans

beberlei:
Blog Post

### DIFF
--- a/source/blog/2023-10-11-orm3-beta1-dbal4-rc1-future-plans.md
+++ b/source/blog/2023-10-11-orm3-beta1-dbal4-rc1-future-plans.md
@@ -30,9 +30,10 @@ This means that we will be making ORM 2.x work with newer versions of PHP,
 fixing security bugs, and introducing layers and features that help with
 forward compatibility in the upgrade path to ORM 3. 
 
-Current users of ORM 2 are strongly encouraged to stay on version 2, as we are
-still working on replacement APIs and forward compatibility, and do not intend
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

source/blog/2023-10-11-orm3-beta1-dbal4-rc1-future-plans.md
```diff
@@ -0,0 +1,65 @@
...
+years, possibly longer, to give you enough time to upgrade and us more time to
+learn from upgrader feedback and improve forward compatibility.
+
+This means that we will be making ORM 2.x work with newer versions of PHP,
+fixing security bugs, and introducing layers and features that help with
+forward compatibility in the upgrade path to ORM 3. 
+
+Current users of ORM 2 are strongly encouraged to stay on version 2, as we are
+still working on replacement APIs and forward compatibility, and do not intend
+to ship them all with ORM 3.0, but with later versions.
```
derrabus, line 35: I would rephrase that as this sounds like we're actively holding back people that want to try the new major release. And if nobody dares to upgrade, we probably won't get valuable feedback on the upgrade path and the removed features.

The release is flagged as beta, which should tell people that they shouldn't use it in production yet. And since we're saying that we're going to maintain ORM 2 for some time, I'd rather emphasize that there's no urgency to upgrade just yet.

derrabus (changes requested):

PR #548, opened 2023-10-24
Allow only node 20 versions

SenseException:
During installation of the website @jwage had issues to install and build it with node 21.  I'm not sure if every 20.x version will work on our website build, but for now we can reduce the node version at least to 20.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
         "webpack-merge-and-include-globally": "^2.3.4"
     },
     "engines": {
-        "node": ">=20"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
         "webpack-merge-and-include-globally": "^2.3.4"
     },
     "engines": {
-        "node": ">=20"
+        "node": ">=20 <21"
```
derrabus, line 35:
```suggestion
        "node": "^20"
```

PR #555, opened 2023-11-21
Realign of the website navigation elements for mobiles

### LINKED ISSUE #337: Navbar glitchy when expanding in a certain viewport size
See it yourself:
Peek 19-02-2020 15-28

This only happens at a specific viewport size.

To reproduce:
- Start from a very small window size and slowly increase its width until the navigation sidebar appears to the left. That's probably the `md` breakpoint.
- Click the hamburguer menu icon to expand the collapsible navbar.

My browser is Firefox 73. Ubuntu 19.10 with GNOME on Xorg session.

SenseException:
I did some work and tests on the website navigation for the usage on mobiles and how we can improve it.

# Current design

Auswahl_001

While the navigation looks good for usage in a regular browser, the mobile view doesn't provide the same experience. In its "closed" state it provides all space for the website content, which is great:

Auswahl_004

Once the navigation gets triggered with the hamburger icon, we currently get this:

Auswahl_005

 Because of this I propose this PR with the following changes to the mobile view of the navigation.

# New design

When the navigation is closed, the search box is already visible and can be used right after entering the website.

Auswahl_002

This allows more space for the search results, especially on smaller devices like a smartphone. Once the navigation gets activated, we're going to get this view:

Auswahl_003

The Doctrine logo doesn't increase the height of the navigation anymore and won't take much of the content space just like the search box currently does. And because the navigation only needs to get activated for going to another page, it won't be needed that often as before, when the search box was part of the collapsed DOM.

As a bonus: This PR will also implicitly fix #337.

I've tested this on
* Firefox (119)
* Chromium (119.0.6045.159)

What isn't handled in this PR:

* The edit button
  * The mobile view probably doesn't need it 
* The sub-navigation
  * Especially the "Projects" nav element takes a lot of space

I'll take on these points on another time.

### DIFF
--- a/source/styles/index.scss
+++ b/source/styles/index.scss
@@ -528,6 +528,17 @@ div.jsactive pre {
   h4 {
     font-size: 10px;
   }
+
+  .doctrine-navbar .nav-outbound {
+    position: absolute;
+    top: 0;
+    right: 0;
+    z-index: 101;
+  }
+
+  #search-box {
+    width: 160px;
+  }
 }
 
 @media (min-width: 576px) and (max-width: 768px) {
--- a/source/styles/index.scss
+++ b/source/styles/index.scss
@@ -74,6 +74,7 @@ h6 {
   background: linear-gradient(-90deg, #2c374c, #1f2e4b);
   color: white;
   padding: 0 10px 0 10px;
+  align-items: normal;
 }
 
 .doctrine-navbar input {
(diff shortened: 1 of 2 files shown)


jwage (approved):

greg0ire (approved):

SenseException:
@stof Could you create a screenshot to show that you describe? The search box has the same size as the desktop version.

SenseException:
I've watched it on one of my older smartphones and I think I get it now. I'll look into it.

SenseException:
@stof I've changed the size of the search box for the smaller devices, but not in a flex form because of the dropdown menues that pop up and cause a similar width problem. In my current changes I introduced layer as a kind of workaround so that the dropdowns won't influence the search box.

I've put the latest changes on staging.

derrabus (approved):

PR #556, opened 2023-11-29
Raise phpstan error level to 8

SenseException:


lib/Docs/RST/RSTCopier.php
```diff
@@ -157,11 +159,18 @@ private function getSidebarRST(string $docsDir): string
...
-            $sidebar = preg_replace('/^([ \t]*)(reference|tutorials|cookbook)/m', '$1/$2', $sidebar);
-
-            return $sidebar;
+            return $this->pregReplace('/^([ \t]*)(reference|tutorials|cookbook)/m', '$1/$2', $sidebar);
         }
 
         return self::DEFAULT_SIDEBAR;
     }
+
+    private function pregReplace(string $pattern, string $replacement, string $content): string
```
SenseException, line 168: It's weird to create a method to make phpstan happy.

SenseException: It's worth a try.

SenseException: I used the Safe PHP library to test the preg_replace. We trade the ` Parameter ... of function preg_replace expects array|string, string|null given.` error with `should return string but returns array<string>|string.`, but not in the same amount of errors as before.

greg0ire: On the one hand, it doesn't look to bad, and I think the new issue could be fixed by contributing generics if really it's an issue, or by wrapping that function in our own function that would only accept strings.

On the other hand, the last merged MR on safe was merged on April 13th… so I'm not sure how, well, _safe_ it is to rely on that library. I'd like to know what other maintainers think.

composer.json
```diff
@@ -48,7 +49,8 @@
     "config": {
         "allow-plugins": {
             "dealerdirect/phpcodesniffer-composer-installer": true,
-            "php-http/discovery": true
+            "php-http/discovery": true,
+            "infection/extension-installer": true
```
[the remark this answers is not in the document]
SenseException: It's weird. I've looked into the package and found nothing about this plugin. When I removed thecodingmachine/safe after a composer self-update and reinstalled it, the plugin wasn't re-added anymore. :thinking: 

The plugin is gone now.

lib/Docs/RST/RSTCopier.php
SenseException: We could either create our own safe preg_replace or create a class with the functionality to return a string. I doubt that we will be going to level 9 without multiple ignored-errors, so I don't see that we'll need more safe functions in the coming weeks. And it would be one less lib in the dependencies.

The pattern in preg_replace can be an array too, so we would be able to reduce the usage of that function and therefore less asserts and I wonder if the order of method invocations, together with str_replace is relevant. Maybe we can reduce the usage of preg_replace in this class.

SenseException:
I've combined the preg_replace calls to reduce the need to call assert on every pattern that gets replaced.

greg0ire (approved):

lib/Commands/BuildAllCommand.php
```diff
@@ -97,7 +97,10 @@ private function runCommand(string $command, array $arguments): int
     {
         $input = new ArrayInput(array_merge(['command' => $command], $arguments));
 
-        return $this->getApplication()->find($command)
-            ->run($input, new ConsoleOutput());
+        $command = $this->getApplication()?->find($command);
+
+        assert($command instanceof Command);
+
+        return $command->run($input, new ConsoleOutput());
```
derrabus, line 104: If `find()` returns, it will always return a command. The assertion looks a bit more specific as it needs to be. The critical part is the `getApplication()` call which might return `null`, so let's only cover that particular case.

```suggestion
        $application = $this->getApplication();

        assert($application !== null);

        return $application->find($command)->run($input, new ConsoleOutput());
```

derrabus (approved):

PR #559, opened 2024-01-09
Fetch team members from GitHub API

SenseException:
This PR introduces fetching the Doctrine team members from the GitHub API. 

The team member and contributor logic seem to have methods and properties that aren't used anymore on the website, but here focus only on the team members API access. Cleanup work will be done in a separate PR.


greg0ire (approved):

alcaeus:
Using doctrine/core is a start, but it will lose us the info on what projects everybody maintains. To get that info, you'd have to iterate through the [Maintainers child teams](https://github.com/orgs/doctrine/teams/maintainers/teams) to see what maintenance team each person is in.

SenseException:
While the YAML config contained this information too, I didn't find where this was used on the website. It could be (re)implemented if that's the case.

alcaeus (approved):

PR #564, opened 2024-02-03
Blog post for Doctrine ORM 3.0 and DBAL 4.0

jwage:


SenseException:
Building the website data step `project_contributors` seems to run too long. I've checked it and it seems that the API call (with GitHub client) to the statistics returns an empty array from the response, which results in even longer `sleep()` cycles inside the loop https://github.com/doctrine/doctrine-website/blob/master/lib/Github/ProdGithubProjectContributors.php#L97. Other API calls look fine though.

jwage:
> Building the website data step `project_contributors` seems to run too long. I've checked it and it seems that the API call (with GitHub client) to the statistics returns an empty array from the response, which results in even longer `sleep()` cycles inside the loop https://github.com/doctrine/doctrine-website/blob/master/lib/Github/ProdGithubProjectContributors.php#L97. Other API calls look fine though.

Should we throw an exception when it returns an empty array so it doesn't run forever?

As for why it is returning an empty array, that is odd. Could it be a temporary api issue?

SenseException:
Somehow the while-loop with `sleep()` is run through by the annotations-repository. These are the current sleep values:

1. 1 second
2. 2 seconds
3. 6 seconds
4. 24 seconds
5. 120 seconds
6. 720 seconds

If I remove `sleep()` and just look at the last iteration, it is going to trigger one with  1401602636313 seconds. :scream:

SenseException:
> Should we throw an exception when it returns an empty array so it doesn't run forever?

It does throw an exception after the last try. Which seem to be 1401602636313 seconds. It will break the build.

> As for why it is returning an empty array, that is odd. Could it be a temporary api issue?

Other API calls are fine, so I wonder why only the statistics are affected.

jwage:
It shouldn't ever be an empty array, right? Or were the retries added due to it failing and returning an empty array sometimes?

SenseException:
Yes, the current code retries in case of an empty array.

https://github.com/doctrine/doctrine-website/blob/master/lib/Github/ProdGithubProjectContributors.php#L54-L56

jwage:
> https://github.com/doctrine/doctrine-website/blob/master/lib/Github/ProdGithubProjectContributors.php#L54-L56

Should we just disable this functionality for now so builds will flow again? and then we can investigate why it is not working or either drop the functionality, fix it, or figure out a new method to replace the existing implementation?

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+
+We are thrilled to announce the release of Doctrine ORM 3.0 and DBAL 4.0.
+These releases are the culmination of over a decade of hard work across
+dozens of contributors and the Doctrine maintainers.
+
+## What's New
+
+**A Slimmer, More Efficient ORM**: The new Doctrine ORM 3.0 comes in at
+326KB, down from 400KB in ORM 2.18.0. This reduction not only makes the
+ORM lighter but also signals our efforts to streamline and optimize every
```
beberlei, line 16: I'd like to add something along the lines:

> And focus our maintenance efforts on the core functionality of an ORM and less on tooling and helpers that are only useful by a small number of our users.

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+**A Growing Community**: The Doctrine project now boasts 1029 contributors
+across all its projects. This vibrant community is the backbone of Doctrine,
+providing valuable insights, feedback, and contributions that drive the
+project forward.
+
+## Upgrading
+
+We understand that upgrading to a new major version can be difficult. The
+best way to upgrade is to first upgrade to the latest Doctrine ORM 2.x and
+DBAL 3.x version and address any deprecation warnings that are reported.
```
beberlei, line 38: Since we have a special way of doing deprecations we should link to where that is explained

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+In addition to that, we've maintained comprehensive documentation about
+every change, deprecation and BC break to facilitate a smooth transition
+to ORM 3.0 and DBAL 4.0.
+
+- [Upgrade to Doctrine ORM 3.0](https://github.com/doctrine/orm/blob/3.0.x/UPGRADE.md)
+- [Upgrade to Doctrine DBAL 4.0](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md)
+
+## Doctrine ORM 2 Future
+
+We plan to maintain Doctrine ORM 2 for the next 2 years by providing bug and security fixes. We may also add or deprecate things in 2.x to improve the
```
beberlei, line 51: "At least"

beberlei (changes requested):

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+
+**Enhanced Code Quality and Coverage**: With ORM 3.0, we've pushed our
+code coverage from 84% to 89%. This improvement underscores our commitment
+to reliability and the stability of the Doctrine ecosystem, ensuring that
+your applications run smoothly.
+
+**Leaner Dependencies**: In Doctrine ORM 3.0, we have finally eliminated
+dependencies on `doctrine/cache` and `doctrine/common`. This change reduces
+complexity and improves maintainability of Doctrine as we now depend on other
+well maintained libraries for caching responsibilities.
```
greg0ire, line 27: Maybe we can even mention we depend on a higher-level standard? There was nothing wrong with `doctrine/cache`, but implementing a PSR means we are more interoperable/easy to use by frameworks.

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
greg0ire: This? https://www.doctrine-project.org/policies/deprecation.html

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+every change, deprecation and BC break to facilitate a smooth transition
+to ORM 3.0 and DBAL 4.0.
+
+- [Upgrade to Doctrine ORM 3.0](https://github.com/doctrine/orm/blob/3.0.x/UPGRADE.md)
+- [Upgrade to Doctrine DBAL 4.0](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md)
+
+## Doctrine ORM 2 Future
+
+We plan to maintain Doctrine ORM 2 for the next 2 years by providing bug and security fixes. We may also add or deprecate things in 2.x to improve the
+existing forward-compatbility layer to make the transition to ORM 3 smoother.
```
greg0ire, line 52: @beberlei should we mention partial hydration here?

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+
+## What's New
+
+**A Slimmer, More Efficient ORM**: The new Doctrine ORM 3.0 comes in at
+326KB, down from 400KB in ORM 2.18.0. This reduction not only makes the
+ORM lighter but also signals our efforts to streamline and optimize every
+aspect of our library.
+
+**Enhanced Code Quality and Coverage**: With ORM 3.0, we've pushed our
+code coverage from 84% to 89%. This improvement underscores our commitment
```
greg0ire, line 20: That figure is for ORM. For DBAL, it goes from 86% to 94%.

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+to reliability and the stability of the Doctrine ecosystem, ensuring that
+your applications run smoothly.
```
greg0ire, line 22: I think it also shows that what we chose to remove was underused and therefore untested code.

SenseException:
> Should we just disable this functionality for now so builds will flow again?

Reducing retry-time is handled in #565

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
...
+path to upgrade.
+
+In addition to that, we've maintained comprehensive documentation about
+every change, deprecation and BC break to facilitate a smooth transition
+to ORM 3.0 and DBAL 4.0.
+
+- [Upgrade to Doctrine ORM 3.0](https://github.com/doctrine/orm/blob/3.0.x/UPGRADE.md)
+- [Upgrade to Doctrine DBAL 4.0](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md)
+
+## Doctrine ORM 2 Future
```
derrabus, line 49:
```suggestion
## The Future of Doctrine ORM 2
```

derrabus (commented):
I like it!

source/blog/2024-02-03-doctrine-orm-3-and-dbal-4-released.md
```diff
@@ -0,0 +1,61 @@
+---
+title: "Doctrine ORM 3 and DBAL 4 Released"
+authorName: Jonathan H. Wage
+authorEmail: jwage@doctrine-project.org
+permalink: /2024/02/03/doctrine-orm-3-and-dbal-4-released.html
+---
+
+We are thrilled to announce the release of Doctrine ORM 3.0 and DBAL 4.0.
+These releases are the culmination of over a decade of hard work across
+dozens of contributors and the Doctrine maintainers.
```
morozov, line 10: As trivia and self-reflection (we don't have to include this in the post). Even though DBAL 3.0.0 was released in 2020, the work on DBAL 4 technically started in 2016 (https://github.com/doctrine/dbal/pull/2499, https://github.com/doctrine/dbal/commit/36522aa4562293c9df54497444992f0a89f1d0f5). So some of the improvements there are also almost 8 years old.

PR #588, opened 2024-06-28
Replace skeleton mapper with orm for sitemappage

SenseException:
Depends on https://github.com/doctrine/rst-parser/pull/281 to be merged. Otherwise duplicate values will break the website build.


greg0ire (approved):
It's merged, but don't you need:
- a release
- to bump the version number, to be on the safe side?

SenseException:
No need to. The website config file of a repository only need to be in the default branch. It wasn't something that needed a new version of a package that gets installed with composer.

PR #589, opened 2024-07-03
Replace skeleton mapper with orm for blogposts

SenseException:
With this PR it looks like data-builder aren't needed anymore. Blogposts are one of two models that are using data-builder for writing data into a cache, that won't be needed when ORM is used.


source/blog/2007-10-09-first-website.md
```diff
@@ -3,7 +3,7 @@ title: "New Website"
 authorName: jwage
 authorEmail:
 categories: []
-permalink: /2007/10/09/new-website.html
+permalink: /2007/10/09/first-website.html
```
SenseException, line 6: I had to rename the permalink because "new-website" is a slug that exists two times for blogs and won't work as ids otherwise without rewriting too much. Because this is an older (the first) blog post, it shouldn't have much of an impact on search engines.

greg0ire (approved):

PR #594, opened 2024-07-12
Introduce phpdoc guides builder

jaapio:
I introduced an interface to be able to switch between the different implementations. The first setup of guides is done at the required locations and allows us to render the documentation. However this doesn't work yet without issues. More might be needed consider this work in progress.

- [x] introduce new code highlighter
- [x] Check why tocs do not contain all items like before
- [x] Add support for toc and tocheader directives
- [x] Add template for toc 
- [x] Review existing directives, and add new ones for the ones that are not supportes by guides.
- [x] check what is wrong with the code highlighter in orm
- [x] Recreate search index using new nodes


jaapio:
I started adding tasks for the things I want to do in this PR, will add more when I find issues.

jaapio:
@SenseException  could you please give me some direction on how to continue with this. I do see that tests are needed and some static analysis issues must be resolved. My first intention was to keep the old doc rendering working if needed I can restore that but I needed to remove some logic in the copy actions because I replaced the way the sidebar menu is moved. 

I'm sure this change will break the search as I didn't address that yet and we could also remove all old rst-parser related code to make this a fresh restart. But I'm not sure we should cover all that in this PR. 

Please let me know what you think.

lib/Commands/BuildAllCommand.php
```diff
@@ -73,7 +72,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         assert(is_bool($clearBuildCache));
 
         if ($clearBuildCache) {
-            array_unshift($commands, 'clear-build-cache');
+            $commands = ['clear-build-cache' => [], ...$commands];
```
jaapio, line 75: seems to be a bug here, that I solved in this pr when passing the clear-build-cache option.

lib/Docs/RST/Logger.php
```diff
@@ -0,0 +1,22 @@
...
+use function sprintf;
+
+use const PHP_EOL;
+
+class Logger extends AbstractLogger
+{
+    /** @param mixed[] $context */
+    public function log(mixed $level, Stringable|string $message, array $context = []): void
+    {
+        echo sprintf('[%s] %s: %s', date('Y-m-d H:i:s'), $level, $message) . PHP_EOL;
```
jaapio, line 20: Yeah... A logger but what should we do? Will anyone ever look at this?

templates/guides/body/admonition.html.twig
```diff
@@ -0,0 +1,30 @@
...
+{% endif %}
+{% if name == 'warning' or name == 'caution' %}
+    {% set admonitionClass='warning' %}
+    {% set textClass='text-primary' %}
+    {% set icon='fa-exclamation-circle' %}
+{% endif %}
+{% if name == 'tip' or name == 'hint' %}
+{% endif %}
+{% if name == 'seealso' %}
+{% endif %}
```
greg0ire, line 16: Are these 2 blocks missing some calls to `set`?

config/packages/guides.yml
```diff
@@ -0,0 +1,14 @@
...
+  themes:
+    doctrine: '%doctrine.website.root_dir%/templates/guides'
+
+code:
+  aliases:
+      html+php: php
+      html+jinja: html
+      php-annotations: php
+      php-attributes: php
+  languages:
```
greg0ire, line 13: What does this do?

lib/Docs/RST/Logger.php
greg0ire: Sorry, where is that logger used? Why is it necessary?

lib/Docs/RST/DoctrineRSTBuilder.php
```diff
@@ -0,0 +1,38 @@
...
+            $directory,
+            $targetDirectory,
+        );
+    }
+
+    /** @return DocumentNode[] */
+    #[Override]
+    public function getDocuments(): array
+    {
+        return $this->builder->getDocuments()->getAll();
```
greg0ire, line 36: I agree with static analysis… (why) is this OK? Do both classes maybe have a similar interface?

lib/Docs/RST/Logger.php
SenseException: I assume it appears in the console output when the build is run.

lib/Docs/RST/RSTPostBuildProcessor.php
```diff
@@ -37,6 +37,8 @@ public function __construct(
 
     public function postRstBuild(Project $project, ProjectVersion $version, RSTLanguage $language): void
     {
+        return;
```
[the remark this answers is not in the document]
SenseException: Does this involve deleting old classes that aren't used anymore?

lib/Twig/MainExtension.php
```diff
@@ -117,11 +119,13 @@ private function getAssetCacheBuster(string $path, string $rootPath): string
...
     }
 
-    public function yamlEncode(mixed $value)
+    public function yamlEncode(mixed $value): mixed
     {
         if (is_string($value)) {
             return str_replace('\\', '\\\\', $value);
         }
+
+        return $value;
```
SenseException, line 128: How did this work in the past when return wasn't present before? Is this method even called?

config/config.yml
```diff
@@ -13,7 +13,7 @@ parameters:
 
     doctrine.website.title: Doctrine
     doctrine.website.subtitle: PHP Open Source Project
-    doctrine.website.url: https://lcl.doctrine-project.org
+    doctrine.website.url: http://lcl.doctrine-project.org
```
SenseException, line 16: This seems to be a misconfiguration and is supposed to be `https://www.doctrine-project.org`. These config values need to be overridden in `config/local.yml`

lib/Docs/RST/RSTCopier.php
```diff
@@ -56,6 +53,10 @@ public function copyRst(Project $project, ProjectVersion $version): void
             $files = $this->rstFileRepository->getSourceFiles($language->getPath());
 
             foreach ($files as $file) {
+                if (substr($file, strlen($language->getPath()) + 1) === 'sidebar.rst') {
```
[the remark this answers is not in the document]
SenseException: Could this be handled with `str_ends_with`?

SenseException:
I've left some comments in the code rather doing a first review. Some (probably now outdated) tests are failing and the static analysis issues need to be addressed.

lib/Docs/RST/RSTPostBuildProcessor.php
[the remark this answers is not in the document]
jaapio: Yes it does, this class should be removed entirely

lib/Docs/RST/Logger.php
jaapio: yes this does the same. we could attach a normal PSR logger

lib/Docs/RST/DoctrineRSTBuilder.php
jaapio: I think the question is, should we invest time on this. Phpstan is right, this is not correct. But I would rather remove this whole class than make it work. As we will move away from the old parser. 

This will be a point of no return when I start the cleanup. This was a first attempt to do things step by step. Which failed.

templates/guides/body/admonition.html.twig
jaapio: Could be, that these are needed. I need to check if the documentation contains those types.

jaapio (commented):
Some responses added

greg0ire:
Suggestion: target `staging` instead of `master`. That way, if you'd rather work with several PRs, we can do that without impacting the real website, and we won't block other PRs on `master` while doing so.

lib/Docs/RST/RSTCopier.php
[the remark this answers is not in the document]
jaapio: the way I do it right now basically to check the full path. Maybe the comparison should be:

`$file === $language->path() . '/sidebar.rst'`

config/config.yml
jaapio: nice spotted, reverted this change

config/packages/guides.yml
jaapio: good catch, I removed part of the config, this config is used by the guides di extensions for symfony DI, to configure the library and all services.

lib/Docs/RST/RSTCopier.php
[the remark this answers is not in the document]
SenseException: I find that suggested line much more readable. Please use it in your if-block. :+1:

lib/Docs/RST/Logger.php
SenseException: This should be sufficient for this PR. A logger can be added later.

lib/Docs/RST/DoctrineRSTBuilder.php
SenseException: Please be aware that I won't approve PRs that do have failing workflows.

jaapio:
I decided to include the cleanup of the old code because that made more sense. I will have a look at the pipeline later on.

SenseException:
You'll also need to rebase your branch to the latest changes that were merged to master.

greg0ire:
I just updated the `staging` branch to make it up to date with `master`, BTW :)

jaapio:
Ok guys, this seems to be green and passing. I'm really looking forward to your experience and review. Let me know if there are any questions!

lib/Docs/SearchIndexer.php
```diff
@@ -63,23 +62,20 @@ public function buildSearchIndexes(
...
-            $this->buildDocumentSearchRecords($document, $records, $project, $version);
+            foreach ($this->buildDocumentSearchRecords($document, $project, $version) as $record) {
+                $records[] = $record;
+            }
         }
 
         $this->getSearchIndex()->saveObjects($records, ['autoGenerateObjectIDIfNotExist' => true]);
     }
 
-    /** @param mixed[][] $records */
```
greg0ire, line 72: I'd still document what is generated.

`/** Generator<mixed[]> */`

lib/Guides/Compiler/SidebarTransformer.php
```diff
@@ -0,0 +1,49 @@
...
+
+use Override;
+use phpDocumentor\Guides\Compiler\CompilerContext;
+use phpDocumentor\Guides\Compiler\NodeTransformer;
+use phpDocumentor\Guides\Nodes\Menu\TocNode;
+use phpDocumentor\Guides\Nodes\Node;
+use phpDocumentor\Guides\RestructuredText\Nodes\GeneralDirectiveNode;
+
+/** @implements NodeTransformer<TocNode|GeneralDirectiveNode> */
+final class SidebarTransformer implements NodeTransformer
```
greg0ire, line 15: What's the purpose of this class? It would be great to have some comments on classes that don't have an obvious purpose such as this one.

greg0ire:
Very few comments on my end, this looks good. What are the next steps? We merge this, then review the staging website, and work on fixing whatever is broken?

lib/Docs/RST/RSTCopier.php
```diff
@@ -106,7 +101,7 @@ private function prepareRSTTemplate(string $content, string $sidebar, string $so
 
         // append the source file name to the content so we can parse it back out
         // for use in the build process
```
SenseException, line 103: This comment is a remnant if the code you removed in the next line. This can be removed too.

lib/Guides/StaticWebsiteGenerator/GuidesRstConverter.php
```diff
@@ -0,0 +1,54 @@
...
+    #[Override]
+    public function getExtensions(): array
+    {
+        return ['rst'];
+    }
+
+    #[Override]
+    public function convertSourceFile(SourceFile $sourceFile): string
+    {
+        $this->themeManager->useTheme('doctrine');
```
SenseException, line 39: Shouldn't the theme come from the guide.yml config?

templates/guides/structure/document.html.twig
```diff
@@ -0,0 +1,29 @@
+---
+title: "{{ node.pageTitle | yaml_encode }}"
```
SenseException, line 2: What's the purpose of this yaml encoding?

phpstan.neon.dist
```diff
@@ -13,11 +13,5 @@ parameters:
         message: "#While loop condition is always true#"
         path: "lib/Commands/BuildWebsiteCommand.php"
         count: 1
-      - # Temporary final classes in tests are covered by nunomaduro/mock-final-classes until the refactorings start
-        message: "#has unresolvable native type#"
-        path: "tests/"
-      - # Temporary final classes in tests are covered by nunomaduro/mock-final-classes until the refactorings start
-        message: "#contains unresolvable type#"
-        path: "tests/"
```
SenseException, line 21: Looks like you can remove the nunomaduro/mock-final-classes dev dependency from composer.json now, right? :tada:

templates/guides/structure/document.html.twig
```diff
@@ -0,0 +1,29 @@
...
+docsIndex: {% if node.filePath ends with 'index' %}true{% else %}false{% endif %}
+
+docsSourcePath: "/en/{{ node.filePath }}.rst"
```
SenseException, line 5: Needs to be tested for old Doctrine 1 once we have this in the stage branch.

templates/guides/website-document.html.twig
```diff
@@ -0,0 +1,13 @@
+{% verbatim %}
+{% block content %}{% verbatim %}
+{% endverbatim %}
+    {% for child in node.children -%}
+        {% if child is not tocNode %}
+        {{ renderNode(child) }}
+        {% endif %}
+    {%~ endfor -%}
+
+{{ "{% endverbatim %}" }}
```
SenseException, line 10: What is this output for?

tests/Twig/MainExtensionTest.php
```diff
@@ -74,6 +74,6 @@ public function testGetAssetUrl(): void
             'http://lcl.doctrine-project.org',
         );
 
-        self::assertSame('http://lcl.doctrine-project.org/js/main.js?de1272', $url);
+        self::assertStringStartsWith('http://lcl.doctrine-project.org/js/main.js', $url);
```
SenseException, line 77: Maybe you can use a regex for the assertion in this test. We expect an assert hash here.

SenseException (changes requested):
> Very few comments on my end, this looks good. What are the next steps? We merge this, then review the staging website, and work on fixing whatever is broken?

`stage` isn't part of a process, but more a trigger and test-branch if you want to share changes that more people are supposed to look into. Since stage became a target branch for this PR, after all comments were addressed, it would be merged to stage and we would test the documentation and "fix whatever is broken" as you already mentioned it.

AFAIK is the search indexer the only thing we currently cannot test on stage.

jaapio:
To answer some of the questions asked in general. We need to have a look at the result of the docs parser, which are twig templates. That's why you see yaml and twig tags in the output and templates. Those templates are picked up by the website build command and processed into the actual page. 

This wasn't my idea but how it currently works. Given the fact that the docs templates do not have information about the other projects and pages I do understand this approach. But it results into very interesting template code. 
The yaml for example is used in the search index building. And some other locations. We can improve this, but I think that's out of scope for now.

lib/Docs/SearchIndexer.php
jaapio: I added phpstan types to document the array shapes

jaapio: I added phpstan types to document the array shapes

lib/Guides/StaticWebsiteGenerator/GuidesRstConverter.php
jaapio: I introduced a compiler pass that will set the theme

greg0ire (approved):
Great work 👍

SenseException (approved):
Let's see how this looks on staging :+1:

jaapio:
Thanks for your reviews!

PR #605, opened 2024-08-21
Avoid copying sidebar

greg0ire:
It should not be built as a document.

Extracted from https://github.com/doctrine/doctrine-website/pull/594

### DIFF
--- a/lib/Docs/RST/RSTCopier.php
+++ b/lib/Docs/RST/RSTCopier.php
@@ -56,6 +56,10 @@ class RSTCopier
             $files = $this->rstFileRepository->getSourceFiles($language->getPath());
 
             foreach ($files as $file) {
+                if ($file === $language->getPath() . '/sidebar.rst') {
+                    continue;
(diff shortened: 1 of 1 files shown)


lib/Docs/RST/RSTCopier.php
```diff
@@ -56,6 +57,10 @@ public function copyRst(Project $project, ProjectVersion $version): void
             $files = $this->rstFileRepository->getSourceFiles($language->getPath());
 
             foreach ($files as $file) {
+                if (str_ends_with($file, 'sidebar.rst')) {
```
jaapio, line 60: please be aware that this might result into all files with `sidebar.rst` to be skipped. a file like `who-to-render-sidebar.rst` will also not be copied. 

That's why my initial implementation was looking so odd.

greg0ire: Fixed 👍

jaapio (approved):

PR #609, opened 2024-09-09
Update PHPStan to 1.12

SenseException:


lib/Docs/RST/RSTPostBuildProcessor.php
```diff
@@ -80,6 +81,8 @@ private function processHtmlFile(
         // parse out the source file that generated this file
         preg_match('/<p>{{ DOCS_SOURCE_PATH : (.*) }}<\/p>/', $contents, $match);
 
+        assert(count($match) === 2);
```
derrabus, line 84: Asserting that the regex did actually match should also narrow the match array enough.

```suggestion
        $result = preg_match('/<p>{{ DOCS_SOURCE_PATH : (.*) }}<\/p>/', $contents, $match);

        assert($result === 1);
```

SenseException: It does. I didn't expect that phpstan can work with the return value to handle the matches.

SenseException:
It was a PR from the future, but I downgraded the version to the one that is known to the current time. 😇

derrabus (approved):

PR #613, opened 2024-09-24
Fix section author markup

jaapio:
Apply the correct template for section authors.
Add an extra fix in the migration from old doc styles to correct the display of author names.

See: https://staging.doctrine-project.org/projects/doctrine-orm/en/3.3/cookbook/implementing-arrayaccess-for-domain-objects.html#implementing-arrayaccess-for-domain-objects

### DIFF
--- a/lib/Docs/RST/RSTCopier.php
+++ b/lib/Docs/RST/RSTCopier.php
@@ -112,6 +112,14 @@ class RSTCopier
         // get rid of .. include:: toc.rst
         $content = str_replace('.. include:: toc.rst', '', $content);
 
+        $content = preg_replace(
+            '/.. sectionauthor:: ([^\(]+) \(([^\)]+)\)/',
+            '.. sectionauthor:: $1 <$2>',
+            $content,
+        );
+
+        assert(is_string($content));
+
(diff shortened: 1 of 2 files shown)


lib/Docs/RST/RSTCopier.php
```diff
@@ -112,6 +112,14 @@ private function fixRSTSyntax(Project $project, string $content): string
         // get rid of .. include:: toc.rst
         $content = str_replace('.. include:: toc.rst', '', $content);
 
+        $content = preg_replace(
+            '/.. sectionauthor:: ([^\(]+) \(([^\)]+)\)/',
+            '.. sectionauthor:: $1 <$2>',
+            $content,
+        );
```
greg0ire, line 119: Why is this necessary? Are there docs that don't follow the desired syntax already?

greg0ire (approved):

lib/Docs/RST/RSTCopier.php
jaapio: Yes, some of the old docs to use `jaapio (jaap@example.com)` where it should be `jaapio <jaap@example.com>`

SenseException (approved):

PR #614, opened 2024-09-24
Replace logger with monolog

jaapio:
The logger introduced before was very usefull in the setup process but by now, it should be replaced by something more mature to ensure the logs to not end up in the final output.


greg0ire (approved):

SenseException:
This needs a rebase and a conflict needs to be fixed due to one of the past PRs introduced changes to composer.lock. Has anyone used the "Resolve conflicts" button of GitHub? I was quite tempted to use it, but I rather prefer to do such steps in a console.

derrabus (approved):

greg0ire:
I remember trying it in the past, but I usually stick to the console, and would certainly do so for `composer.lock`. That being said, there's a chance @jaapio allows writing to his branch, so you could try fixing it locally. I'll give it a try.

greg0ire:
Yup, writable, that's usually the case. I use https://ldez.github.io/prm/ for this, but I think `gh` also covers this feature.

PR #615, opened 2024-09-24
Full wildcard match on default menu

jaapio:
We do want to include every document in the menu when we do a fallback to the default menu. `/*` does match all files are root level, while we wanted to match everything `*`.


greg0ire (approved):

SenseException (approved):
I'm not sure about the extend of this change, so let's see how the menus look like.

PR #616, opened 2024-09-25
Render menu titles

jaapio:
image


SenseException (approved):
From the commit message:

> The upgrade of guides contains the fix required to make the menu
render the section titles of each page.

PR #617, opened 2024-09-25
Improve menu generation

jaapio:
So the origin of the menu's not working is in the way the menu's are defined. The doctrine parser did use * as a deep search wildcard which is not according to the behavior sphinx does it. As the toctree is a directive defined by sphinx we try to follow that one closely. The correct way of writing a toctree is

```
.. toctree::
   :glob:

   **
```

This will include all documents and sections in a project. starting at the level of the current page.

### DIFF
--- a/lib/Guides/Compiler/GlobMenuFixerTransformer.php
+++ b/lib/Guides/Compiler/GlobMenuFixerTransformer.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Website\Guides\Compiler;
+
+use phpDocumentor\Guides\Compiler\CompilerContext;
+use phpDocumentor\Guides\Compiler\NodeTransformer;
+use phpDocumentor\Guides\Nodes\Menu\GlobMenuEntryNode;
+use phpDocumentor\Guides\Nodes\Node;
+
+/**
+ * This class will fix the glob menu entries.
+ *
+ * The glob menu entries are not correctly added in the docs.
+ * This transformer will fix the glob menu entries by replacing the '*' with '**'. This add a certain
+ * risk if the user had the intention to match only documents in the current directory. But this can be solved
+ * adding a `/` at the beginning of the glob pattern.
+ *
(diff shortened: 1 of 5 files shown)


lib/Guides/Compiler/GlobMenuFixerTransformer.php
```diff
@@ -0,0 +1,51 @@
...
+ * This class will fix the glob menu entries.
+ *
+ * The glob menu entries are not correctly added in the docs.
+ * This transformer will fix the glob menu entries by replacing the '*' with '**'. This add a certain
+ * risk if the user had the intention to match only documents in the current directory. But this can be solved
+ * adding a `/` at the beginning of the glob pattern.
+ *
+ * @implements NodeTransformer<GlobMenuEntryNode>
+ */
+final class GlobMenuFixerTransformer implements NodeTransformer
```
greg0ire, line 22: If I understand correctly, we are not following the rst specification properly in some parts of the documentation. If that's the case, it would be nice to have a logging mechanism to this so that we ultimately fix the broken docs and remove this transformer. I think the same could also be done for the thing you did yesterday about the syntax of `section-author`

greg0ire (approved):

lib/Guides/Compiler/GlobMenuFixerTransformer.php
jaapio: Question is, are you going to fix all versions in history?

greg0ire: Oh right, I missed that very important point!

SenseException: That is true, but I started to drop the rendering of some older versions (currently [DBAL](https://github.com/doctrine/doctrine-website/blob/master/config/data/projects.yml#L24) and [ORM](https://github.com/doctrine/doctrine-website/blob/master/config/data/projects.yml#L36) only), so we could reach a point in the projects' history where all issues can be fixed.

greg0ire:
Thanks @jaapio !

PR #622, opened 2024-10-04
Render menu captions when provided

greg0ire:
In my previous PR, I added CSS style for an element that does not get rendered, presumably because of this template override.


linawolf (approved):
Looks good to me

jaapio (approved):

SenseException (approved):

PR #629, opened 2024-10-16
fix: copy more than just rst files

greg0ire:
Now that we use `literalinclude`, source files include more than just rst files. I'd say anything with an extension goes.

This addresses an issue with guides not finding files referenced in `literalinclude` directives, because they were never copied.


SenseException:
Build work fine, but I get this result in the docs:

Auswahl_002

Annotations, PHP, etc. appears below that restructured text, which is not visible in the screenshot.

SenseException (approved):
Latest changes worked with ORM's docs and literalinclude

PR #631, opened 2024-10-22
Remove unneeded autofixes

jaapio:
- `:depth` and `:maxdepth` are both supported by guides
- term text role was never used in all docs
- `code::` is an alias of `code-block`


SenseException (approved):
Yep, I couldn't find `:term:` either.

PR #632, opened 2024-10-22
Use guides to parse and render markdown

jaapio:
This is an effort to remove the static website generator project as a dependency. In it's current state guides-markdown does not support all we need to replace the markdown rendering, so I'm working on that while fixing this project as well. 

- https://github.com/phpDocumentor/guides/issues/1160

With linkable headings: 

image

And nice code blocks like the docs have:

image


SenseException:
Nice. I'll test this when I get the time. Can you please add a test for the newly introduced class?

SenseException (changes requested):
Looks great. Code snippets look better now and headlines have anchors. I found a few anomalies in the rendering of markdown pages though:

## Ordered and unordered lists gaps

Because of a `<p>` tag that is rendered with the `<li>` and `<ol>` ones do lists have a gap of a paragraph for all markdown pages.

(https://www.doctrine-project.org/2016/01/05/dbal-2-5-4-and-2-4-5.html)
Auswahl_006

## Missing links and anchor 

In at least one case aren't the links in a headline rendered and the anchor can't be used. The anchor also looks wrong being dark and bigger than on other pages. I assume this might have something to do with the brackets in the headline.

(https://www.doctrine-project.org/2021/11/26/dbal-3.2.0.html)
Auswahl_005

jaapio:
I checked the headlines in the blogpost that look malformed, and those are actaully links within the headline. And that makes it look very strange. 

```
## Support for `psr/log` ([#4967](https://github.com/doctrine/dbal/pull/4967))
``` 

As the headline is rendered as a link, and includes a link.... yeah that's where things get really ugly... 
Honestly I don't know how to solve that, do you want me to update those posts? And transform the link to the issue to a link directly below the headline?

jaapio:
The issue with the paragraphs is covered in https://github.com/phpDocumentor/guides/pull/1168

jaapio:
Fixed the list:

image

jaapio:
Although I do agree with you tests are very much needed the added class in this PR is very hard to test. Basically we just glue some external parts together. Adding a test would be nothing more than a very complicated mock and it will basically test nothing as the parser is an external project and the noderenderer is as well. 

The rest will be covered by static analysis.

SenseException:
> do you want me to update those posts? And transform the link to the issue to a link directly below the headline?

Yes, that would probably be the best.

> Adding a test would be nothing more than a very complicated mock

There are also functional tests, like in `tests/FunctionalTest.php` where a service can be used. The content for the rendering itself can be controlled with the injected `Doctrine\StaticWebsiteGenerator\SourceFile\SourceFile` object.

jaapio:
I did check this pr again and it looks like the new added code is being tested by the functional tests already. The functional tests are rendering the full source directory, and some files are validated by the tests. The blog post `/2018/04/06/new-website.html` is one of the checked files. So I think this PR is complete now.

jaapio:
I just did a rebase, pipeline is green

SenseException (approved):

SenseException:
Thanks, @jaapio

PR #650, opened 2025-01-30
Require dependencies mentioned in the code

greg0ire:
This should make it easier to understand weird and sudden build issues.


derrabus:
Why does this fix the issue? The lockfile did not change except for the hash.

greg0ire:
It will prevent accidentally upgrading Flysystem when only meaning to upgrade `phpDocumentor/guides`, which depends on it. Also, there may be a way to configure dependabot to open separate PRs for major ugprades, if that's not the default behavior already.

jaapio:
I think this is an issue on my side, I will have a look tonight

jaapio:
I just checked, this is not an issue on my side. I did release a new version yesterday which allows flysystem v3 as well using a separate package named `phpdocumentor/filesystem`. This package is optional and should not be required. The extra packages you added are needed by were already installed. By requiring them directly you might prevent an accidental upgrade. 

For this particular case just adding flysystem ^1.0 would have been sufficient. 

So I think this PR is good to be merged.

greg0ire:
> For this particular case just adding flysystem ^1.0 would have been sufficient.

Yes, I thought I would run `composer-require-checker` and make sure each dependency in the output is covered. I should have mentioned that.

greg0ire:
@derrabus I finally understood the issue and why I don't reproduce it locally. We reuse the static analysis workflow, and it does a `composer update`, but this is not a library, and `composer install` should be used here instead.

greg0ire:
The proper fix for the build is https://github.com/doctrine/doctrine-website/pull/654

composer.json
```diff
@@ -13,16 +13,21 @@
...
         "algolia/algoliasearch-client-php": "^3.3",
+        "doctrine/collections": "^2",
         "doctrine/dbal": "^4.0",
         "doctrine/inflector": "^2.0",
         "doctrine/orm": "^3.2",
         "doctrine/static-website-generator": "^2.1",
         "erusev/parsedown": "^1.7",
         "knplabs/github-api": "^3.9",
+        "league/flysystem": "^1",
+        "league/tactician": "^1",
```
SenseException, line 24: I don't understand why this is supposed to be a dependency.

greg0ire: @SenseException because it is mentioned here: https://github.com/doctrine/doctrine-website/blob/69a03f9b49164a937a24f8f8b544196a2a0b24d2/lib/Docs/RST/GuidesParser.php#L11

SenseException (approved):
Please rebase your branch

PR #652, opened 2025-01-30
Remove Algolia SDK for PHP

greg0ire:
It could have been removed in #207 .


greg0ire:
My bad, it is still used as a Symfony Service.

PR #715, opened 2025-04-15
Introduce additional sort order for projects

SenseException:
Projects with a higher rate of usage have higher traffic on their docs. This Pr allows to order ORM, DBAL and ODM higher in our project lists on the website making them easier to find for users of the website.

### DIFF
--- a/lib/Repositories/ProjectRepository.php
+++ b/lib/Repositories/ProjectRepository.php
@@ -44,7 +44,7 @@ class ProjectRepository extends EntityRepository
         return $this->findBy([
             'active' => true,
             'integration' => false,
-        ], ['name' => 'asc']);
+        ], ['sortOrder' => 'asc', 'name' => 'asc']);
     }
 
     /** @return Project[] */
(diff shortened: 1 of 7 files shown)


lib/Repositories/ProjectRepository.php
```diff
@@ -44,7 +44,7 @@ public function findPrimaryProjects(): array
         return $this->findBy([
             'active' => true,
             'integration' => false,
-        ], ['name' => 'asc']);
+        ], ['sortOrder' => 'asc', 'name' => 'asc']);
```
beberlei, line 47: Shouldnt that be:

```suggestion
        ], ['sortOrder' => 'desc', 'name' => 'asc']);
```

SenseException: That would move ORM, DBAL and ODM to the end of the project's list in the navigation and homepage. It would be the opposite of what this PR tries to accomplish.

greg0ire (approved):

PR #717, opened 2025-04-24
Php 8.4

SenseException:


phpcs.xml.dist
```diff
@@ -9,7 +9,7 @@
     <!-- Ignore warnings, show progress of the run and show sniff names -->
     <arg value="nps"/>
 
-    <config name="php_version" value="80300"/>
+    <config name="php_version" value="80400"/>
```
greg0ire, line 12: Why is this line even needed? This is no library.

SenseException: True. I removed it.

greg0ire (approved):

Issue #571, opened 2024-02-29
Git command in contributor guide broken


derrabus:
On https://www.doctrine-project.org/contribute/index.html

<img width="503" alt="Bildschirmfoto 2024-02-29 um 10 45 17" src="https://github.com/doctrine/doctrine-website/assets/1506493/78b63b4c-f7c9-403c-b564-33fa8ce45d15">

Source:

https://github.com/doctrine/doctrine-website/blob/65820892a2e36266f8cb953ef0a4886a159d9a8b/source/contribute.rst?plain=1#L65-L68

greg0ire:
Thank you 🙏

PR #748, opened 2025-06-17
Fix git commands and links

osavchenko:
Relates to https://github.com/doctrine/doctrine-website/issues/571

During the fix, I noticed that the "Maintainer Workflow" link is broken, and the "Security Policy" link can be unified with others.

### DIFF
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -64,8 +64,8 @@ ambiguous, we will add two explicit names for remote Git repositories:
 
 .. code-block:: console
     $ [`cd` into the directory created by `git clone ...`]
-    $ git remote add doctrine git@github.com:doctrine/<name-of-project>.git
-    $ git remote add fork git@github.com:<your-github-name>/<name-of-project>.git
+    $ git remote add doctrine git@github.com:doctrine/&lt;name-of-project&gt;.git
+    $ git remote add fork git@github.com:&lt;your-github-name&gt;/&lt;name-of-project&gt;.git
 
 Choosing the right branch
 -------------------------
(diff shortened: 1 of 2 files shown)


source/contribute.rst
```diff
@@ -64,8 +64,8 @@ ambiguous, we will add two explicit names for remote Git repositories:
 
 .. code-block:: console
     $ [`cd` into the directory created by `git clone ...`]
-    $ git remote add doctrine git@github.com:doctrine/<name-of-project>.git
-    $ git remote add fork git@github.com:<your-github-name>/<name-of-project>.git
+    $ git remote add doctrine git@github.com:doctrine/&lt;name-of-project&gt;.git
+    $ git remote add fork git@github.com:&lt;your-github-name&gt;/&lt;name-of-project&gt;.git
```
greg0ire, line 68: @jaapio hi :wave: 

This is one of the buggy pages: https://www.doctrine-project.org/contribute/index.html

I think this is more like a rendering bug, wouldn't you agree? I don't think the source rst should be changed.

jaapio: Yes this looks like a rendering bug. The code blocks should have been escaped. And apparently they are not... Which is a security issue?

greg0ire: Thanks for looking into it. They don't seem to do any HTML escaping indeed… but how would we replace these classes? Do you already have something built-in in phpdocumentor for rendering this kind of thing?

jaapio: https://github.com/doctrine/doctrine-website/pull/749 

Here you go, it's still in draft as I need to finish some missing things like aliases and it needs to be tested.

greg0ire: Oh wow thanks!

SenseException:
This ticket depends on #749 afaiu

osavchenko:
@greg0ire, sorry for the long response. Rebased.

greg0ire (approved):

SenseException:
Thank you @osavchenko

PR #749, opened 2025-06-18
Resolve issue with code rendering

jaapio:
The current implementation is vulnerable for cross site scripting. The risk is low as the documentation is maintained and reviewed by the doctrine team. 

This new implementation removes some old code, and replaces that with components from phpdocumentor/guides.  The new implementation is using css based line numbers which allow users to select the text like they did before, the only missing feature are links to particular line numbers. 

image

image


source/js/main.js
```diff
@@ -34,17 +34,15 @@ export default function () {
...
-      var element = document.createElement('textarea');
-      element.value = copyText;
-      document.body.appendChild(element);
-      element.select();
-      document.execCommand('copy');
-      document.body.removeChild(element);
+      try {
+        await navigator.clipboard.writeText(copyText);
+      } catch (error) {
+        console.error(error.message);
```
SenseException, line 44: I believe you can throw that error in case the clipboard doesn't work. `console` may not be available in all browser and can cause errors itself.

SenseException (changes requested):
When I click the Copy-To-Clipboard icon, the clipboard contains `undefined` instead of a code string.

jaapio:
> When I click the Copy-To-Clipboard icon, the clipboard contains undefined instead of a code string. 

Sorry about that, I could not test it local as the clipboard is only available on https. Now I fixed my local setup, and tested it local. It does work with the new javascript I added.

SenseException (changes requested):
Firefox:

```
Uncaught (in promise) TypeError: navigator.clipboard is undefined
    __WEBPACK_DEFAULT_EXPORT__ webpack://doctrine-website/./source/js/main.js?:60
```

Chrome:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'writeText')
    at HTMLButtonElement.eval (main.js:60:33)
```

jaapio:
The clipboard is only available on https. Did you test this local on http?

SenseException:
You're right, it was on http. I've triggered another staging deployment so we can test it there again.

jaapio:
Checked on my phone, using chrome and Firefox. Both work.

Thanks for deploying

SenseException (approved):
I can confirm that the copy script works again. Thank you for the improvements.

PR #818, opened 2025-10-07
Fix rendering of rst examples

SenseException:
The RST examples page wasn't rendering properly anymore and also was rendering RST that was supposed to be shown as RST code.

I introduced RST code examples plus how they are looking like when rendered.

### DIFF
--- a/source/rst-examples.rst
+++ b/source/rst-examples.rst
@@ -1,11 +1,24 @@
 RST Examples
 ============
 
-.. sectionauthor:: Jonathan H. Wage <jonwage@gmail.com>
+Versions
+--------
 
-..
+Code
+~~~~
 
-Jonathan H. Wage wrote these examples!
+.. code-block:: rst
+
+    .. versionadded:: 2.2
+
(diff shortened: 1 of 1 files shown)


source/rst-examples.rst
```diff
@@ -1,11 +1,24 @@
 RST Examples
 ============
 
-.. sectionauthor:: Jonathan H. Wage <jonwage@gmail.com>
```
SenseException, line 4: This doesn't do anything to the rendered doc file.

source/rst-examples.rst
```diff
@@ -15,21 +28,89 @@ Jonathan H. Wage wrote these examples!
...
-.. versionadded:: 2.2
+Code Examples
+-------------
 
-This was added new in version 2.2.
+Code
+~~~~
 
-.. code-block:: php
+.. code-block:: rst
```
SenseException, line 37: `rst` currently doesn't do anything because it isn't a valid code-highlighting on the Doctrine RST rendering. This can be introduced to the rendering later though.

